PR #25, opened 2024-06-14
Crypto

Rhaki:
Implements new tests for signatures and new hash functions.


crates/wasm/src/imports.rs
```diff
@@ -36,7 +36,12 @@ extern "C" {
     // return value of 0 means ok; any value other than 0 means error.
     fn secp256r1_verify(msg_hash_ptr: usize, sig_ptr: usize, pk_ptr: usize) -> i32;
     fn secp256k1_verify(msg_hash_ptr: usize, sig_ptr: usize, pk_ptr: usize) -> i32;
-    fn secp256k1_pubkey_recover(msg_hash_ptr: usize, sig_ptr: usize, recovery_id: u8) -> usize;
+    fn secp256k1_pubkey_recover(
+        msg_hash_ptr: usize,
+        sig_ptr: usize,
+        recovery_id: u8,
+        compressed: u8,
```
larry0x, line 43: Using a `bool` for `compressed` doesn't work?

Rhaki: bool shouldn't be supported in export "C"

crates/crypto/src/secp256k1.rs
```diff
@@ -1,20 +1,33 @@
...
+    // low-S only, this step was not needed.
+    if let Some(normalized) = sig.normalize_s() {
+        sig = normalized;
+    }
+
+    validate_length!(
+        pk,
+        SECP256K1_PUBKEY_COMPRESSED_LEN,
+        SECP256K1_PUBKEY_UNCOMPRESSES_LEN
+    )?;
```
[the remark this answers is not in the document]
Rhaki: Yes, it's only to handle the error. It can be removed. Anyway the signature Error is always empty. It should be taken in consideration to properly emit a custom error depending on the failing reason

crates/crypto/src/hashers.rs
```diff
@@ -0,0 +1,75 @@
...
+#[cfg(test)]
+mod tests {
+    use hex_literal::hex;
+
+    macro_rules! hash_test {
+        ($hash_fn:ident, $world:expr => $hexed:expr) => {
+            #[test]
+            fn $hash_fn() {
+                let hash = super::$hash_fn($world.as_bytes());
+                assert_eq!(hash, hex!($hexed))
```
larry0x, line 60: Use `hex_literal::hex!` which decodes the HEX strings at compile time. Also avoid casting `hash` to a `Vec`.

PR #28, opened 2024-06-19
Index map

larry0x:


crates/storage/src/index_map.rs
```diff
@@ -0,0 +1,451 @@
...
+            name_surname: MultiIndex::new(
+                |_, data| (data.name.clone(), data.surname.clone()),
+                "pk_namespace",
+                "name_surname",
+            ),
+            id: UniqueIndex::new(|data| data.id, "unique_ns"),
+        };
+
+        IndexedMap::<u64, Foo, FooIndexes<'a, E>, E>::new("pk_namespace", indexes)
+    }
```
larry0x, line 255: `IndexedMap`s can be constants, so it's unnecessary to create them using a function:

```rust
const FOO: IndexedMap<u64, Foo, FooIndexes> = IndexedMap::new("pk_namespace", FooIndexes {
    name: MultiIndex::new(|_, data| data.name.clone(), "pk_namespace", "name"),
    name_surname: MultiIndex::new(
        |_, data| (data.name.clone(), data.surname.clone()),
        "pk_namespace",
        "name_surname",
    ),
    id: UniqueIndex::new(|data| data.id, "unique_ns"),
});
```

crates/macros/src/lib.rs
```diff
@@ -111,3 +115,34 @@ pub fn grug_export(_attr: TokenStream, mut item: TokenStream) -> TokenStream {
...
+            quote! { &self.#name }
+        })
+        .collect::<Vec<_>>();
+
+    let expanded = quote! {
+        #input
+
+        impl cw_storage_plus::IndexList<#ty> for #struct_ty<'_> {
+            fn get_indexes(&'_ self) -> Box<dyn Iterator<Item = &'_ dyn cw_storage_plus::Index<#ty>> + '_> {
+                let v: Vec<&dyn cw_storage_plus::Index<#ty>> = vec![#(#names),*];
```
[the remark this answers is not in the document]
Rhaki: Yeah. This macro has to be rewrite also because now the Indexes struct could have different Encoding schema.
See [here](https://github.com/left-curve/grug/blob/fd81a9030b317d8e2a142d3a4f300acd43a9dad5/crates/storage/src/index_map.rs#L230C1-L241)

crates/storage/src/index_map.rs
Rhaki: In this example we need `E: Encoding` to properly test with different encoding schema

crates/macros/src/lib.rs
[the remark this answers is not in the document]
Rhaki: Even in a real-world environment, it makes no sense to define an `IndexMap` where the `Encoding` is changeable. In the contract, once decided, the schema will always be the same. So in the `macro`, we can omit the encoding and define the `IndexMap`as constant, where the `Encoding` must already be defined

crates/storage/src/index_map.rs
```diff
@@ -0,0 +1,430 @@
...
+impl<'a, K, T, I, E: Encoding<T>> IndexedMap<'a, K, T, I, E>
+where
+    K: MapKey + Clone,
+    I: IndexList<T>,
+    T: Clone,
+    E: Encoding<T>,
+{
+    pub fn update<A, Err>(&'a self, storage: &mut dyn Storage, key: K, action: A) -> Result<T, Err>
+    where
+        A: FnOnce(Option<T>) -> Result<T, Err>,
```
larry0x, line 185: We changed the `update` function signature from CosmWasm's. Now the `action` function should a `Result<Option<T>, Err>`. Returning a `None` means the data is to be deleted.

crates/storage/src/index_map.rs
larry0x: > In this example we need `E: Encoding` to properly test with different encoding schema

Understood

PR #33, opened 2024-07-01
Gas metering and wasm module cache

larry0x:


crates/vm/wasm/src/vm.rs
```diff
@@ -33,17 +46,29 @@ impl Vm for WasmVm {
         storage: StorageProvider,
         querier: QuerierProvider<Self>,
         code: &[u8],
+        gas_tracker: SharedGasTracker,
     ) -> VmResult<WasmInstance> {
+        let code_hash = hash(code);
```
larry0x, line 51: TODO: `App` should provide the code hash, so that we don't re-do the hash here

crates/app/src/app.rs
```diff
@@ -355,33 +391,56 @@ where
...
+            gas_tracker.clone(),
             tx.sender.clone(),
             msg.clone(),
         )?);
+
+        #[cfg(feature = "tracing")]
+        {
+            let consumed = gas_tracker.read_access().used() - before_consumed;
+            debug!("Gas used: {consumed}");
+        }
```
larry0x, line 412: We can put tracing in `GasTracker` instead

crates/vm/wasm/src/vm.rs
larry0x: We can fix this in a follow up PR, to keep the diff of this one smaller

larry0x:
TODO: gas consumption for:
- setting config
- upload code
- wasm imports

PR #41, opened 2024-07-04
Handling of gas depletion

larry0x:
Previously we're able to handle the case where an import function consumes too much gas, but we can't handle it when the contract does too many operations. This PR fixes this.


crates/vm/wasm/src/environment.rs
```diff
@@ -164,10 +165,20 @@ impl Environment {
...
-            .map_err(Into::into)
+            .map_err(|runtime_err| {
+                // The call has failed. Now we need see why - did gas run out or
+                // other reasons?
+                match get_remaining_points(store, instance) {
+                    // Gas wasn't depleted. It was some other reasons
+                    MeteringPoints::Remaining(_) => VmError::Runtime(runtime_err),
+                    // Gas was depleted. Throw a gas-specific error.
+                    MeteringPoints::Exhausted => VmError::GasDepletion,
+                }
```
larry0x, line 181: Here's the key

crates/vm/wasm/src/vm.rs
```diff
@@ -108,6 +116,10 @@ impl Vm for WasmVm {
         let instance = wasmer::Instance::new(&mut store, &module, &import_obj)?;
         let instance = Box::new(instance);
 
+        // Set gas limit on the metering
+        let gas_remaining = gas_tracker.remaining().unwrap_or(u64::MAX);
+        set_remaining_points(&mut store, &instance, gas_remaining);
```
larry0x, line 121: Here's another key step

Issue #40, opened 2024-07-04
Wasm instance memory limit


larry0x:
CosmWasm sets memory limit for each Wasm instance as follows. We should do the same:

```rust
let tunables = LimitingTunables::new(base, limit_to_pages(limit));
engine.set_tunables(tunables);
```

larry0x:
Completed by #149

PR #49, opened 2024-07-09
for #40

dzmitry-lahoda:
for #40


larry0x:
Thanks @dzmitry-lahoda, but we have completed this in #149.

PR #54, opened 2024-07-10
ADR-070 unordered transactions

Rhaki:
The implementation is not done at the chain (`App`) level but is entirely managed by the `Account` contract.

A field `data` of type `serde_json::Value` has been added to the struct `Tx` where extra data can be included.

The account contract looks for the `account` key in this field and attempts to deserialize its content into `AccountData`, which contains information about whether the transaction is ordered or not.

If ordered, the message to be verified with the public key is created by including the sequence. If unordered, the `expiration` serialized in JSON is inserted instead of the sequence.

The `expiration` with the hashed message is then inserted into a `Set` (there are actually two sets, one with the key `(timestamp, Hash)` and the other with `(height, Hash)`). If the key is already present or if the expiration has already expired, the transaction is rejected.

With each call to the `authenticate_tx` function, expired unordered transactions are removed from the two `Set`s.


larry0x:
Since contracts in this repo are just for testing purpose, and not for production use, I suggest we keep them simple and don't implement this here.

We can implement this in a separate, production contracts repo.

PR #58, opened 2024-07-16
Code pinning

larry0x:


crates/vm/wasm/src/cache.rs
```diff
@@ -32,18 +34,26 @@ impl Cache {
...
+        self.modules
             .write_access()
             .put(code_hash.clone(), (module.clone(), engine.clone()));
 
         Ok((module, engine))
     }
+
+    pub fn try_remove_from_modules(&self, code_hash: &Hash) {
+        self.modules.write_access().pop(code_hash);
+    }
```
larry0x, line 58: This should be unnecessary if you put `self.modules` and `self.pinned` under a single `Shared`.

crates/vm/wasm/src/vm.rs
```diff
@@ -147,6 +129,67 @@ impl Vm for WasmVm {
...
+                }
+
+                for i in new_pinned {
+                    // Remove from default cache
+                    self.cache.try_remove_from_modules(&i);
+
+                    let code = CODES.load(storage, &i)?;
+                    // Signlepass still here, need to be changed with LLVM
+                    let mut compiler = Singlepass::new();
+                    compiler.canonicalize_nans(true);
```
larry0x, line 159: Should we put this in `build_from_compiler`?

crates/vm/wasm/src/vm.rs
```diff
@@ -147,6 +129,67 @@ impl Vm for WasmVm {
...
+            })?;
+
+        Ok(())
+    }
+}
+
+pub fn build_from_compiler<C: CompilerConfig + Into<Engine>>(
+    mut compiler: C,
+    code: &[u8],
+) -> VmResult<(Module, Engine)> {
```
larry0x, line 174: I suggested using the `with` syntax for all trait bounds throughout our codebase, for consistency. It'a bit more verbose, but more readable imo.

```rust
fn build_from_compiler<C>(mut compiler: C, code: &[u8]) -> VmResult<(Module, Engine)>
where
    C: CompilerConfig + Into<Engine>,
{
    // ...
}
```

Also, does this function needs to be public?

larry0x:
> Still this PR make sense? using LLVM would not make grug compatible with zk. We could use Cranelift for pinned and Singlepass for non pinned, or using Cranelift for every contract

Yeah, closing this PR. See: https://github.com/left-curve/grug/issues/57#issuecomment-2237782627

PR #62, opened 2024-07-18
Prefix range

larry0x:


crates/storage/src/bound.rs
```diff
@@ -17,16 +21,20 @@ pub enum RawBound {
 pub enum Bound<K> {
     Inclusive(K),
     Exclusive(K),
-    InclusiveRaw(Vec<u8>),
-    ExclusiveRaw(Vec<u8>),
```
larry0x, line 21: We never use these two variants, so removed.

crates/storage/src/bound.rs
```diff
@@ -17,16 +21,20 @@ pub enum RawBound {
...
 }
 
 impl<K> Bound<K> {
-    pub fn inclusive<T: Into<K>>(t: T) -> Self {
+    pub fn inclusive<T>(t: T) -> Self
+    where
+        T: Into<K>,
```
larry0x, line 29: From now, I want to use the `where` syntax everywhere in the codebase, for consistency.

crates/storage/src/indexed.rs
```diff
@@ -70,7 +65,7 @@ where
     where
         T: 'b,
     {
-        self.no_prefix().range_raw(store, min, max, order)
+        self.primary.range_raw(store, min, max, order)
```
larry0x, line 68: Instead of creating a `self.no_prefix` method and working with the `Prefix`, we now simply delegate to `self.primary`.

crates/storage/src/key.rs
```diff
@@ -48,23 +53,20 @@ pub trait Key {
...
-    type Prefix: Key;
+    type Prefix: Prefixer;
 
     /// For tuple keys, the elements _excluding_ the `Prefix`.
     ///
     /// E.g. for `(A, B)`, this is `B`.
     ///
     /// Use `()` for singleton keys.
-    type Suffix: Key;
+    type Suffix;
```
larry0x, line 63: This trait bound is unnecessary

crates/storage/src/key.rs
```diff
@@ -48,23 +53,20 @@ pub trait Key {
...
 
     /// The type that raw keys deserialize into, which may be different from the
     /// key itself.
     ///
     /// E.g. when `&str` is used as the key, it deserializes into `String`.
-    ///
-    /// Must be an _owned_ type, as denoted by the `'static` lifetime requirement.
-    /// In comparison, the key itself is almost always a reference type.
-    type Output: 'static;
+    type Output;
```
larry0x, line 69: Same as above, unnecessary.

crates/storage/src/key.rs
```diff
@@ -346,7 +365,97 @@ macro_rules! impl_integer_key {
     }
 }
 
-impl_integer_key!(i8, u8, i16, u16, i32, u32, i64, u64, i128, u128);
+impl_integer_key!(
+    u8, u16, u32, u64, u128, i8, i16, i32, i64, i128, Uint64, Uint128, Uint256, Uint512,
+);
```
larry0x, line 370: Implement `Key` trait for `Uint{64,128,256,512}`.

Can't do the same for `Int`, `Udec`, `Dec` because they don't implement `Bytable`.

PR #65, opened 2024-07-19
Add a bunch of new methods to TestSuite

larry0x:


.github/workflows/rust.yml
```diff
@@ -16,7 +16,7 @@ jobs:
           toolchain: stable
 
       - name: Run tests
-        run: cargo test --all-targets
+        run: cargo test
```
larry0x, line 19: It turns out benchmarks are also ran if we use the `--all-targets` flag, so remove.

Issue #73, opened 2024-07-22
CLI and SDK improvements


larry0x:
1. `grug_sdk::SigningKey::create_and_sign_tx` should take a gas limit parameter
2. `grug tx` command should take a `--gas-wanted` flag
3. `grug q tx` command should take the hash in either upper or lowercase (currently only taking lowercase)
4. Avoid printing tracing during `Query` and `CheckTx`

PR #74, opened 2024-07-22
Update CheckTx; improvements to the SDK and the CLI

larry0x:
Closes: #73

### DIFF
--- a/bin/src/query.rs
+++ b/bin/src/query.rs
@@ -101,7 +101,7 @@ enum SubCmd {
     },
     /// Get transaction by hash
     Tx {
-        /// Transaction hash
+        /// Transaction hash in hex encoding
         hash: String,
     },
     /// Get block by height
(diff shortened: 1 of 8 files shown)


larry0x:
Currently, in `CheckTx` we simulate the entire transaction, not just the authentication part. 

This means 1) bigger load on the nodes, and 2) txs that would fail simply won't enter the mempool.

Need to think about whether this is a desired design.

Rhaki:
A transition that fails during the simulation (in this case in the check tx) does not imply that it will also fail during the block insertion. For example, an arbitrage action (which takes into account the txes already in the mempool) could fail in the check tx, but not during the finalize block. Furthermore, if the signature is also verified, there cannot be more than one tx per account in the mempool, as they would have to have two different sequences, and the check tx in the second one would fail because the sequence is not updated

Rhaki:
The second problem could be solved using unordered tx, but it is still a big limitation not being able to have two txes in mempool from the same sender

larry0x:
> A transition that fails during the simulation (in this case in the check tx) does not imply that it will also fail during the block insertion.

True. The converse is also true - succeeding in simulation but can also fail in block.

larry0x:
> if the signature is also verified, there cannot be more than one tx per account in the mempool

Also true, unless we implement ADR-070

larry0x:
Now we only do the authentication part (`before_tx`) in `CheckTx`, instead of the full tx.

Rhaki:
> Now we only do the authentication part (`before_tx`) in `CheckTx`, instead of the full tx.

Consider to add a Query/dedicated entry point on account type contracts for check_tx. This would allow to verify the sign with a sequence that is different from the +1 of the one saved, allowing to have in the mempool multiple txs from the same sender. See [75](https://github.com/left-curve/grug/issues/75)

larry0x:
> allowing to have in the mempool multiple txs from the same sender

For that I think we just implement ADR-070

PR #87, opened 2024-07-30
Make `after_tx` optional

Rhaki:
Based on https://github.com/left-curve/grug/issues/60.


crates/types/src/response.rs
```diff
@@ -65,6 +65,72 @@ impl Response {
...
+
+    pub fn add_attribute<K, V>(mut self, key: K, value: V) -> Self
+    where
+        K: ToString,
+        V: ToString,
+    {
+        self.response = self.response.add_attribute(key, value);
+        self
+    }
+}
```
larry0x, line 132: Adding these methods so that `AuthResponse` is more ergonomic to use.

crates/testing/tests/account.rs
```diff
@@ -96,3 +99,128 @@ fn check_tx_and_finalize() -> anyhow::Result<()> {
...
+        .should_succeed_and_equal(Uint128::ZERO);
+    suite
+        .query_balance(&accounts["sender"], "ugrug")
+        .should_succeed_and_equal(Uint128::new(50_000));
+    suite
+        .query_balance(&accounts["sender"], "nft/badkids/1")
+        .should_succeed_and_equal(Uint128::ZERO);
+
+    Ok(())
+}
```
larry0x, line 226: Add tests for backrunning.

larry0x (approved):

PR #93, opened 2024-08-04
Shorter addresses

larry0x:
- Change address derivation:

   ```diff
   - address := sha256(deployer | code_hash | salt)
   + address := ripemd160(sha256(deployer | code_hash | salt))
   ```

- Add [ERC-55 checksum](https://eips.ethereum.org/EIPS/eip-55) to `Addr`

- Use uppercase for `Hash`


crates/types/src/address.rs
```diff
@@ -138,15 +150,40 @@ impl From<Addr> for String {
...
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "{}{}", Self::PREFIX, self.0)
+        let mut checksum = [0; 40];
+        // This encodes hex in lowercase.
+        hex::encode_to_slice(&self.0, &mut checksum).unwrap();
+
+        let mut hasher = Keccak256::new();
+        // Note we're hashing the UTF-8 hex string, not the raw bytes.
+        hasher.update(checksum);
+        let hash = hasher.finalize();
```
larry0x, line 167: Here we do the keccak hash in the contract instead of using `ctx.api.keccak_256`. However according to our benchmark, there isn't much performance difference. @Rhaki

PR #94, opened 2024-08-05
Type state builder

Rhaki:
Based on https://github.com/left-curve/grug/issues/91.

`Type state builder` pattern has been implemented `GenesisBuilder` and `TestBuilder`.

While this pattern is really useful in `TestBuilder` (for example, it prevents calling `set_account_code` if an account has already been added via `add_account`, or prevents calling the `set_owner` function if an account has not yet been set), in `GenesisBuilder` it could be a constraint because in order to call the `build` function, it is required to call the `set_bank`, `set_taxman`, `with_upload_permission`, `with_instantiate_permission` functions.
Passing these parameters as arguments in the `new` function would make the flow much clearer, since compilation errors are complex to interpret, they do not help to understand why the `build` method cannot be called clearly


larry0x:
> Passing these parameters as arguments in the new function

The thing is you need to first use the `GenesisBuilder` to deploy those contracts, so that you know their addresses, before you can set them in the config.

Rhaki:
We could add these contracts on build fn as arguments

sdk/rust/src/genesis_builder.rs
```diff
@@ -3,73 +3,153 @@ use {
...
+            bank: self.bank,
+            taxman: self.taxman,
+            cronjobs: self.cronjobs,
+            instantiate_permission: self.instantiate_permission,
+            upload_permission: self.upload_permission,
+            upload_msgs: self.upload_msgs,
+            other_msgs: self.other_msgs,
+        }
     }
+}
```
larry0x, line 63: This ensures that we can only `set_owner` once.

PR #96, opened 2024-08-06
Message::Upload gas consumption

Rhaki:


crates/app/src/providers.rs
```diff
@@ -15,7 +15,7 @@ use {
 #[derive(Clone)]
 pub struct StorageProvider {
     storage: Box<dyn Storage>,
-    namespace: Vec<u8>,
+    pub namespace: Vec<u8>,
```
larry0x, line 18: I think this should be kept private, and instead add a `.namespace()` method that returns a `&[u8]`.

This is because with this public, the namespace can be mutated, which isn't intended:

```rust
storage.namespace = vec![1, 2, 3, 4, 5];
```

crates/app/src/gas/storage.rs
```diff
@@ -58,6 +58,21 @@ where
...
+        key: K,
+    ) -> StdResult<bool>;
+
+    fn save_with_gas(
+        &self,
+        storage: &mut dyn Storage,
+        gas_tracker: GasTracker,
+        key: K,
+        value: &T,
+    ) -> StdResult<()>;
```
larry0x, line 74: Minor styling - let's put `save_with_gas` below `range_with_gas`, because the way I order these is all non-mutable methods first, then all mutable ones.

PR #98, opened 2024-08-09
Storage tests + CryptoError + minor fixes

Rhaki:


contracts/tester/src/types.rs
```diff
@@ -35,4 +35,30 @@ pub enum QueryMsg {
     /// `db_remove`, or `db_remove_range` FFI import methods directly. We need
     /// to test whether the VM can properly reject this behavior.
     ForceWrite { key: String, value: String },
+
+    CryptoVerify {
+        ty: CryptoVerifyType,
+        pk: Vec<u8>,
+        sig: Vec<u8>,
+        msg_hash: Vec<u8>,
```
larry0x, line 43: Don't use `Vec<u8>` in messages because they get serialized into Json arrays, e.g.

```json
{
  "pk": [
    1,
    2,
    3,
    4,
    5
  ]
}
```

Use `Binary` (if length is unknown) or `ByteArray<N>` (if length is known) instead. Here these should be `ByteArray<N>`.

contracts/tester/src/types.rs
```diff
@@ -35,4 +35,30 @@ pub enum QueryMsg {
...
+    },
+    RecoverSepc256k1 {
+        sig: Vec<u8>,
+        msg_hash: Vec<u8>,
+        recovery_id: u8,
+        compressed: bool,
+    },
+
```
larry0x, line 51: Remove unnecessary empty line

crates/crypto/src/error.rs
```diff
@@ -1,27 +1,31 @@
-use thiserror::Error;
+use grug_types::CryptoError;
 
-#[derive(Debug, Error)]
-pub enum CryptoError {
-    #[error(transparent)]
-    Signature(#[from] signature::Error),
+pub type CryptoResult<T> = core::result::Result<T, CryptoError>;
```
larry0x, line 3: Can you put this in grug-types, right below `CryptoError`

crates/crypto/src/error.rs
```diff
@@ -1,27 +1,31 @@
...
 
-    #[error("data is of incorrect length: expecting {expect}, found {actual}")]
-    IncorrectLength { expect: usize, actual: usize },
+pub(crate) trait SignatureResultExt {
+    type Inner;
+    fn crypto_verify_failed(self) -> CryptoResult<Self::Inner>;
+    fn crypto_recovery_failed(self) -> CryptoResult<Self::Inner>;
+    fn crypto_invalid_pk_format(self) -> CryptoResult<Self::Inner>;
+    fn crypto_invalid_sig_format(self) -> CryptoResult<Self::Inner>;
```
larry0x, line 10:
```suggestion
    type Inner;

    fn crypto_verify_failed(self) -> CryptoResult<Self::Inner>;

    fn crypto_recovery_failed(self) -> CryptoResult<Self::Inner>;

    fn crypto_invalid_pk_format(self) -> CryptoResult<Self::Inner>;

    fn crypto_invalid_sig_format(self) -> CryptoResult<Self::Inner>;
```

crates/storage/src/codec.rs
```diff
@@ -42,3 +42,43 @@ where
         from_proto_slice(data)
     }
 }
+
```
larry0x, line 45: Always put this to separate the test from other contents of the file

```suggestion

// ----------------------------------- test ------------------------------------

```

crates/storage/src/counter.rs
```diff
@@ -55,3 +55,43 @@ where
         Ok(new_value)
     }
 }
+
```
larry0x, line 58:
```suggestion

// ----------------------------------- test ------------------------------------

```

crates/storage/src/counter.rs
```diff
@@ -55,3 +55,43 @@ where
...
+#[cfg(test)]
+mod tests {
+
```
larry0x, line 61: Remove this empty line

crates/storage/src/index/unique.rs
```diff
@@ -67,11 +69,45 @@ where
...
         self.idx_map.remove(storage, idx);
     }
 }
+
+#[derive(BorshDeserialize, BorshSerialize, Debug, Clone)]
+pub struct UniqueValue<PK, T> {
+    pub value: T,
+    pk: Vec<u8>,
+    p: PhantomData<PK>,
+}
```
larry0x, line 88: What's the purpose of this?

Rhaki: This is the value used by UniqueIndex internally in the Map to save both data and the PK. The PK is serialized via Key. We can replace this using the a tuple

larry0x: We already have `grug_types::Record`, although it's an alias to `(Vec<u8>, Vec<u8>)`. You can modify it as:

```rust
pub type Record<K = Vec<u8>, V = Vec<u8>> = (K, V);
```

and use it instead?

crates/storage/src/prefix.rs
```diff
@@ -7,17 +7,25 @@ use {
...
-pub struct Prefix<K, T, C>
+pub trait AsKeysIterator {}
+pub trait AsRangeIterator {}
+
+pub struct KeysIterator;
+impl AsKeysIterator for KeysIterator {}
+
+pub struct RangeIterator;
+impl AsRangeIterator for RangeIterator {}
+impl AsKeysIterator for RangeIterator {}
```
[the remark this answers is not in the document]
Rhaki: Similar to the type state builder, these two structures with attached traits allow to change the functions that can be used in the PrefixPath. This is because in the case of the Set is possible to use both the keys and range functions, where range returned a record with the value = Empty (that is not needed at all). When the Set returns the PrefixPath, it creates it with the KeysIterator, while a map with the RangeIterator. The methods for iterating the keys are implemented for both the RangeIterator and the KeysIterator, while those for iterating Range and Values ​​are implemented only for the RangeIterator

crates/storage/src/index/unique.rs
Rhaki: The problem with using Record as the Map value is that we implement Deref for UniqueIndex. So the load method or any iteration function will return a (Vec<u8>, T) value. You will always need to Deserialize Vec<u8> into the PK using Key::from_slice. The UniqueIndex structure instead automates this process. I would suggest leaving it

Rhaki:
Closed. Will split in more PR

PR #100, opened 2024-08-10
Faster `Client::send_messages_with_confirmation`

Rhaki:
In `Client::send_messages_with_confirmation`, the `await` calls for `chain_id`, `sequence`, and `gas_limit` are now executed concurrently using `tokio::try_join`, resulting in a 3x speedup in execution time.


larry0x:
LGTM

larry0x:
@Rhaki I don't think grug-client should depend on tokio, because a user may want to use other async runtimes. Can you switch to [futures](https://crates.io/crates/futures) can use `futures::join`?

Rhaki:
Switched

PR #110, opened 2024-08-14
Better config updates

larry0x:
Introduces a `ConfigUpdates` struct, so that when updating a config it's no longer necessary to input the entire config.


crates/types/src/tx.rs
```diff
@@ -34,8 +36,8 @@ pub enum Message {
     ///
     /// For app-level config, setting a value to `Null` means to delete it.
```
Rhaki, line 37: Remove this comment

larry0x: What's wrong with this comment? If you wish no delete a key in the app config, set it to `Null`. If you don't wish to change it, don't include it in `app_updates` map.

PR #112, opened 2024-08-15
Better counter

larry0x:
- Allow user to specify the starting value and step size for the counter.
- Return both the old and new value in the `Counter::increment` method.
- Add a `Counters` (plural) type which is a wrapper over a `Map`

### DIFF
--- a/crates/storage/src/counter.rs
+++ b/crates/storage/src/counter.rs
@@ -1,57 +1,48 @@
 use {
     crate::{Borsh, Codec, Item},
-    grug_types::{Number, NumberConst, StdResult, Storage},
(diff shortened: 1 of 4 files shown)


crates/storage/src/counter.rs
```diff
@@ -1,57 +1,98 @@
...
+        Self {
+            map: Map::new(storage_key),
+            base,
+            step,
+        }
+    }
+}
+impl<'a, K, T, C> Counters<'a, K, T, C>
+where
+    K: Key + Copy,
```
larry0x, line 77: `K` needs to be `Copy`, but we typically use reference `&T` as keys, which is indeed `Copy`-able.

PR #115, opened 2024-08-16
refactor: merge impl_next! macro into generate_uint!

najeal:
This PR merge the two macros `impl_next!` & `generate_uint!` into `generate_uint!`


larry0x:
Thanks @najeal, but we are currently doing a major refactor of the math crate, and this PR is no longer relevant.

Issue #107, opened 2024-08-14
Add balances changes to simulation


larry0x:
It's be helpful to know coin balances changes when running a simulation.

larry0x:
Not planned. Balance changes can already be found by parsing the events emitted by the bank contract.

PR #117, opened 2024-08-18
feat: Balances tracker

najeal:
This is related to #107 

It adds balances changes to `TxOutcome` message. It currently tracks balances changes for all AuthMode but can be limited to the `AuthMode:Simulate`


larry0x:
Thanks @najeal. Unfortunately, we realized this isn't necessary - when you simulate a transaction, the chain returns a list of events emitted. You can already easily find out all balance changes by parsing the events emitted by the bank contract.

PR #119, opened 2024-08-20
feat: `#[derive(grug::QueryRequest)]`

Rhaki:
`gurg_macro::Query` is a new `proc_macro_derive` that automatize some traits implementation and Struct generation for improve the developer experience for execute query between contracts. 

Usage:
```rust
#[derive(grug::Query)]
#[grug::derive(serde)]
pub enum QueryMsg {
   #[returns(ConfigResponse)
   Config {}
   #[returns(StateResponse)
   State(StateRequest)
}
``` 

`grug::Query` check each enum variant and if the variant is `syn::Fields::Named` (like `Config {}`), a new `Struct` called `ConfigRequest` is created with the same parameters (no one in this case).

```rust
pub struct ConfigRequest {}
```

Then for each `Request` `Struct` the macro impl `impl From<#struct_name> for #name` and the trait `QueryResponseType` for mapping `Request => Response`.

Also QuerierWrapper expose a new method called `query_wasm_super_smart`

```rust
pub fn query_wasm_super_smart<R>(&self, contract: Addr, msg: R) -> StdResult<R::Response>
where
    R: AsQueryMsg + QueryResponseType,
    <R as AsQueryMsg>::QueryMsg: Serialize,
    <R as QueryResponseType>::Response: DeserializeOwned,
{
    let request = msg.into_query_msg();

    self.inner
        .query_chain(QueryRequest::WasmSmart {
            contract,
            msg: to_json_value(&request)?,
        })
        .and_then(|res| from_json_value(res.as_wasm_smart()))
}
```

### DIFF
--- a/crates/macros/src/query.rs
+++ b/crates/macros/src/query.rs
@@ -1,13 +1,26 @@
 use {
     core::panic,
     proc_macro::TokenStream,
-    quote::{quote, ToTokens},
-    syn::{parse_macro_input, Data, DeriveInput, Ident, Type},
+    quote::quote,
+    syn::{parse_macro_input, Data, DeriveInput, Fields, Ident, Type},
 };
 
+/// Throughout the function, we will use comments to illustrate how it works,
+/// based on the following example:
+///
+/// ```rust
+/// #[derive(grug::Query)]
+/// enum QueryMsg {
+///     #[returns(String)]
+///     Foo { bar: u64 },
+///     #[returns(Buzz)]
+///     Fuzz(i128),
+/// }
+/// ```
 pub fn process(input: TokenStream) -> TokenStream {
     let input = parse_macro_input!(input as DeriveInput);
 
+    // E.g. `QueryMsg`
     let name = input.ident;
 
     let Data::Enum(data) = input.data else {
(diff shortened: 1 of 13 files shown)


crates/macros/src/query.rs
```diff
@@ -0,0 +1,123 @@
...
+                        #field_name: val.#field_name,
+                    });
+                }
+
+                // Generate the struct definition
+                let struct_def = quote! {
+                    pub struct #struct_name {
+                        #(#fields_struct_definition)*
+                    }
+                };
```
larry0x, line 72: How about instead of generating a response type for each `QueryMsg` variant, we generate a single `QueryResponse` type?

E.g. for this query message:

```rust
#[derive(grug::Query)]
#[grug::derive(serde)]
enum QueryMsg {
    #[returns(String)]
    Foo {},
    #[returns(u64)]
    Bar {},
}
```

We generate this:

```rust
#[grug::derive(serde)]
enum QueryResponse {
    Foo(String),
    Bar(u64),
}
```

crates/testing/tests/queries.rs
```diff
@@ -52,3 +53,126 @@ fn handling_multi_query() -> anyhow::Result<()> {
...
+
+    pub fn query_host(ctx: ImmutableCtx, msg: QueryMsgHost) -> anyhow::Result<Json> {
+        match msg {
+            QueryMsgHost::Config {} => Ok(to_json_value(&CONFIG.load(ctx.storage)?)?),
+        }
+    }
+
+    // ---- Client ----
+    #[derive(Serialize, Deserialize, grug_macros::Query)]
+    #[query_path_override(::grug_types)]
```
larry0x, line 95: If you put this test in `crates/std/tests/queries.rs`, you can then import the macro as `grug::Query`, there's no need to define this override.

I made a test for `#[grug::index_list]` this way.

crates/macros/src/query.rs
Rhaki: Because the compiler is not able to derive the variant of the enum. The purpose of having a bind of Request => Response is not to guess (or match the Response enum variant)

Rhaki: If the QueryFooRequest return a String, is better to have the String returned by the query. Having and Enum requires to check if the variant is Foo and then take the String

PR #121, opened 2024-08-21
Replace serializer and hasher functions with extension traits

larry0x:
Adds the following new traits:

- `JsonExt`
- `BorshExt`
- `ProtoExt`
- `HashExt`


bin/src/tx.rs
```diff
@@ -117,8 +116,8 @@ impl TxCmd {
                 updates,
                 app_updates,
             } => {
-                let updates = from_json_str(&updates)?;
-                let app_updates = from_json_str(&app_updates)?;
+                let updates = <_>::from_json_str(&updates)?;
+                let app_updates = <_>::from_json_str(&app_updates)?;
```
larry0x, line 120: A drawback of extension trait is we need to explicitly write out the type `T` when calling functions like `T::from_json_str`.

contracts/account/src/execute.rs
```diff
@@ -28,7 +28,7 @@ use {
 /// attacker uses the signature to broadcast another transaction on chain B.)
 pub fn make_sign_bytes<Hasher, const HASH_LEN: usize>(
     hasher: Hasher,
-    msgs: &[Message],
+    msgs: &Vec<Message>,
```
larry0x, line 31: `&[T]` doesn't work for some reason. Have to be `&Vec<Message>`

larry0x:
Closed this in favor of #124 and #125

PR #126, opened 2024-08-21
Changes to make tx broadcasting work

larry0x:
1. Make `{finalize,withhold}_fee` methods take an `AuthCtx` instead of `SudoCtx`
2. Taxman to skip fee deduction during simulations
3. Add the `data` field to `UnsignedTx` (our app account needs this for simulation)
4. Add a `broadcast_tx` method to `grug_client::Client`


contracts/taxman/src/execute.rs
```diff
@@ -66,10 +71,15 @@ pub fn withhold_fee(ctx: SudoCtx, tx: Tx) -> StdResult<Response> {
...
-pub fn finalize_fee(ctx: SudoCtx, tx: Tx, outcome: TxOutcome) -> anyhow::Result<Response> {
+pub fn finalize_fee(ctx: AuthCtx, tx: Tx, outcome: TxOutcome) -> anyhow::Result<Response> {
     let cfg = CONFIG.load(ctx.storage)?;
     let info = ctx.querier.query_info()?;
 
+    // In simulation mode, don't do anything.
+    if ctx.mode == AuthMode::Simulate {
+        return Ok(Response::new());
+    }
+
```
Rhaki, line 82: Why not set 
```rust   
    let fee_rate = if ctx.mode == AuthMode::Simulate {
        Dec128::Zero
   } else {
        config.fee_rate
   }
```
And using `fee_rate` instead of `config.fee_rate` on the next lines?

PR #132, opened 2024-08-24
NF-2: impl Deref for Item

Rhaki:
Based on https://github.com/left-curve/grug/pull/118#issuecomment-2307173322: 3

### DIFF
--- a/crates/storage/src/item.rs
+++ b/crates/storage/src/item.rs
@@ -1,16 +1,15 @@
 use {
     crate::{Borsh, Codec, Path},
-    grug_types::{StdError, StdResult, Storage},
-    std::marker::PhantomData,
+    std::{marker::PhantomData, ops::Deref},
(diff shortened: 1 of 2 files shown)


crates/storage/src/item.rs
```diff
@@ -1,16 +1,15 @@
...
 };
 
 pub struct Item<'a, T, C = Borsh>
 where
     C: Codec<T>,
 {
-    storage_key: &'a [u8],
     data: PhantomData<T>,
     codec: PhantomData<C>,
+    path: Path<'a, T, C>,
```
larry0x, line 12: If we have `Path<'a, T, C>` here then we don't need the two `PhantomData`s any more? If so, they can be removed.

Rhaki: Yes, forgot to remove

PR #140, opened 2024-08-29
ICS-23 proofs

larry0x:
Support ICS-23 merkle proof format, which is necessary for IBC to work.


Cargo.toml
```diff
@@ -26,7 +26,7 @@ members  = [
 version       = "0.0.0"
 authors       = ["Left Curve Software Limited (https://leftcurve.software/)"]
 edition       = "2021"
-rust-version  = "1.73"
+rust-version  = "1.80"
```
larry0x, line 29: We use `LazyLock` which is only available since 1.80.

PR #141, opened 2024-08-29
Storage tests

Rhaki:


crates/storage/src/item.rs
```diff
@@ -28,3 +28,272 @@ impl<'a, T, C: Codec<T>> Deref for Item<'a, T, C> {
...
+    struct Config {
+        pub owner: String,
+        pub max_tokens: i32,
+    }
+
+    const CONFIG: Item<Config> = Item::new("config");
+
+    #[test]
+    fn save_and_load() {
+        let mut store = MockStorage::new();
```
larry0x, line 50: Let's name this `storage` instead of `store`, for consistency

crates/storage/src/map.rs
```diff
@@ -425,3 +423,1304 @@ mod test {
...
+
+        let data_1_raw = data1.to_borsh_vec().unwrap();
+        let data_2_raw = data2.to_borsh_vec().unwrap();
+
+        // let's try to iterate!
+        let all: Vec<_> = PEOPLE
+            .range_raw(&store, None, None, Order::Ascending)
+            .collect();
+
+        assert_eq!(2, all.len());
```
larry0x, line 605: This is redundant because we already compare the two vectors, which including comparing their lengths.

crates/storage/src/map.rs
```diff
@@ -425,3 +423,1304 @@ mod test {
...
+        assert_eq!(all, vec![
+            (b"jim".to_vec(), data_2_raw.clone()),
+            (b"john".to_vec(), data_1_raw.clone())
+        ]);
```
larry0x, line 609: this can just be an array `[T; N]` instead of a `Vec<T>`, because `Vec<T>` implements `Partial<[T; N]>`.

Array is faster because it doesn't involve dynamic allocation.

crates/storage/src/map.rs
```diff
@@ -425,3 +423,1304 @@ mod test {
...
+    fn basic_update() {
+        let mut store = MockStorage::new();
+
+        let add_ten = |a: Option<u64>| -> StdResult<_> { Ok(Some(a.unwrap_or_default() + 10)) };
+
+        // save and load on three keys, one under different owner
+        let key: (&[u8], &[u8]) = (b"owner", b"spender");
+        ALLOWANCE.update(&mut store, key, add_ten).unwrap();
+        let twenty = ALLOWANCE.update(&mut store, key, add_ten).unwrap().unwrap();
+        assert_eq!(20, twenty);
```
larry0x, line 1317: Also for consistency, in `assert_eq!` let's always put the actual value first, expected value second.

crates/storage/src/map.rs
```diff
@@ -425,3 +423,1304 @@ mod test {
...
+                None => Ok(Some(Data {
+                    name: "Newborn".to_string(),
+                    age: 0,
+                })),
+            }
+        };
+
+        let old_john = PEOPLE.update(&mut store, b"john", birthday)?.unwrap();
+        assert_eq!(33, old_john.age);
+        assert_eq!("John", old_john.name.as_str());
```
larry0x, line 1357: You don't need `.as_str()` here, because `String` implements `PartialEq<&str>.`

crates/storage/src/map.rs
```diff
@@ -425,3 +423,1304 @@ mod test {
...
+        assert_eq!("Newborn", new_jack.name.as_str());
+
+        // update also changes the store
+        assert_eq!(old_john, PEOPLE.load(&store, b"john")?);
+        assert_eq!(new_jack, PEOPLE.load(&store, b"jack")?);
+
+        // removing leaves us empty
+        PEOPLE.remove(&mut store, b"john");
+        let empty = PEOPLE.may_load(&store, b"john")?;
+        assert_eq!(None, empty);
```
larry0x, line 1370: Prefer

```rust
assert!(something.is_none());
```

over

```rust
assert_eq!(something, None);
```

crates/storage/src/set.rs
```diff
@@ -150,3 +148,397 @@ where
...
+        DOUBLE.insert(storage, (Dec128::from_str("-2")?, "b"))?;
+        DOUBLE.insert(storage, (Dec128::from_str("-2")?, "c"))?;
+        DOUBLE.insert(storage, (Dec128::from_str("-2")?, "d"))?;
+        DOUBLE.insert(storage, (Dec128::from_str("-2")?, "e"))?;
+        DOUBLE.insert(storage, (Dec128::from_str("-1.5")?, "a"))?;
+        DOUBLE.insert(storage, (Dec128::from_str("-1.5")?, "b"))?;
+        DOUBLE.insert(storage, (Dec128::from_str("0")?, "abcb"))?;
+        DOUBLE.insert(storage, (Dec128::from_str("1")?, "b"))?;
+        DOUBLE.insert(storage, (Dec128::from_str("2")?, "b"))?;
+        DOUBLE.insert(storage, (Dec128::from_str("2")?, "a"))?;
```
larry0x, line 401: For things like this, let's use a `for` loop.

crates/storage/src/set.rs
```diff
@@ -150,3 +148,397 @@ where
...
+                .prefix(Dec128::from_str("-2")?)
+                .keys(storage, None, None, Order::Ascending)
+                .collect::<StdResult<Vec<_>>>()?;
+
+            assert_eq!(val.len(), 5);
+            assert_eq!(val[0], "a".to_string());
+            assert_eq!(val[1], "b".to_string());
+            assert_eq!(val[2], "c".to_string());
+            assert_eq!(val[3], "d".to_string());
+            assert_eq!(val[4], "e".to_string());
```
larry0x, line 415: This can be simplified as just

```rust
assert_eq!(val, ["a", "b", "c", "d", "e"]);
```

Apparently, `Vec<String>` implements `PartialEq<[&str; N]>` which is handy.

crates/storage/src/map.rs
```diff
@@ -425,3 +423,1304 @@ mod test {
         }
     }
 }
+
+#[cfg(test)]
+mod cosmwasm_test {
```
larry0x, line 428: Should be plural ("tests")

PR #142, opened 2024-08-30
Improvements to NonZero

larry0x:
`NonZero::new` method now errors instead of panics in case a zero value is provided.

Also, implement necessary traits such that `NonZero<T>` can be used in messages or storage.

### DIFF
--- a/crates/types/src/bytearray.rs
+++ b/crates/types/src/bytearray.rs
@@ -96,7 +96,7 @@ impl<const N: usize> ser::Serialize for ByteArray<N> {
     where
         S: ser::Serializer,
     {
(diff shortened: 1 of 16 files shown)


crates/types/src/bytearray.rs
```diff
@@ -96,7 +96,7 @@ impl<const N: usize> ser::Serialize for ByteArray<N> {
     where
         S: ser::Serializer,
     {
-        serializer.serialize_some(&BASE64.encode(&self.0))
+        serializer.serialize_str(&BASE64.encode(&self.0))
```
larry0x, line 99: This is unrelated to this `NonZero`, but a bug I spotted while working on this PR

Rhaki (approved):

PR #143, opened 2024-09-02
T-WasmVm-1: Cache

Rhaki:
- Handled Cache with zero capacity:
- Add some unit tests for Cache


larry0x:
@Rhaki See my latest commit: https://github.com/left-curve/grug/pull/143/commits/4f30519b0199f3f18d392386250e301ebfa86b00

1. I feel only asserting the cache length == 1 is insufficient, so I add a `Metrics` to the cache and assert the number of cache hits & misses are also as expected.
2. I realize in the `get_or_build_with` we call `self.inner.write_access()` twice, which is bad. I adjusted the implementation so that the write lock is only requested once.

PR #145, opened 2024-09-03
 C-WasmVm-2: Instance

Rhaki:
`WasmInstace` now proivde a `use_env_mut` function that reduce the verbose in the code for using `& mut env`

```rust
// Old
let mut fe_mut = self.fe.clone().into_mut(&mut self.store);
let (env, mut store) = fe_mut.data_and_store_mut();

// New
self.use_env_mut(|env, store| { 
...
}
```

### DIFF
--- a/crates/vm/wasm/src/vm.rs
+++ b/crates/vm/wasm/src/vm.rs
@@ -30,13 +30,14 @@ const MAX_QUERY_DEPTH: usize = 3;
 
 #[derive(Clone)]
 pub struct WasmVm {
-    cache: Option<Cache>,
+    cache: Cache,
 }
 
 impl WasmVm {
     pub fn new(cache_capacity: usize) -> Self {
+        // TODO: handle the case where cache capacity is zero (which means not to use a cache)
         Self {
-            cache: NonZeroUsize::new(cache_capacity).map(Cache::new),
+            cache: Cache::new(NonZeroUsize::new(cache_capacity).unwrap()),
         }
     }
 }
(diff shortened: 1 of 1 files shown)


crates/vm/wasm/src/vm.rs
```diff
@@ -28,14 +30,13 @@ const MAX_QUERY_DEPTH: usize = 3;
...
-    cache: Cache,
+    cache: Option<Cache>,
 }
 
 impl WasmVm {
     pub fn new(cache_capacity: usize) -> Self {
-        // TODO: handle the case where cache capacity is zero (which means not to use a cache)
         Self {
-            cache: Cache::new(NonZeroUsize::new(cache_capacity).unwrap()),
+            cache: NonZeroUsize::new(cache_capacity).map(Cache::new),
```
larry0x, line 39: It seems changes from #143 related to the cache are also included in this PR.

Probably because you created the branch off the #143 branch instead of main.

Can you somehow rebase it so that changes from #143 aren't included in this PR?

crates/vm/wasm/src/vm.rs
```diff
@@ -175,22 +182,29 @@ pub struct WasmInstance {
     fe: FunctionEnv<Environment>,
 }
 
+impl WasmInstance {
+    fn use_env_mut<T>(&mut self, callback: impl FnOnce(&mut Environment, &mut StoreMut) -> T) -> T {
```
larry0x, line 186: I always like to use the `where` syntax, which would be:

```rust
fn use_env_mut<T, F>(&mut self, callback: F) -> T
where
    F: FnOnce(&mut Environment, &mut StoreMut) -> T,
{
    // ...
}
```

larry0x (commented):
I like this. Just two comments:

Rhaki:
@larry0x applied

larry0x:
> @larry0x applied

Ok, if tests pass we can merge

PR #147, opened 2024-09-03
T-WasmVm-4: Memory

Rhaki:
- added test for memory/Region.
- added a new function `validate_region` for extra checks on Region


Rhaki:
@larry0x suggestions applied

PR #148, opened 2024-09-04
T-WasmWm-5: Enviroment

Rhaki:
Added tests for Enviroment

### DIFF
--- a/crates/vm/wasm/src/environment.rs
+++ b/crates/vm/wasm/src/environment.rs
@@ -273,3 +273,243 @@ impl Environment {
         }
     }
 }
+
+// ----------------------------------- tests -----------------------------------
+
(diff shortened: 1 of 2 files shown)


crates/vm/wasm/src/environment.rs
```diff
@@ -273,3 +273,415 @@ impl Environment {
...
+                gas_checkpoint,
+            );
+
+            env.set_wasmer_memory(&instance).unwrap();
+            env.set_wasmer_instance(&instance).unwrap();
+            env
+        };
+
+        (env, store, instance)
+    }
```
larry0x, line 361: I reorganized this function into sections (one for `store` and `instance`, one for `env`) for better readability.

PR #153, opened 2024-09-11
Improvements to testing framework

larry0x:
Introduce `MockContext`, `MockApi`, `MockQuerier` types for use in unit tests.

Although we encourage using `TestSuite` for integration tests, sometimes unit tests are also helpful.

How to use:

```rust
// Use `with_*` methods to customize the context when creating it.
let mut ctx = MockContext::new()
    .with_chain_id("dev-1")
    .with_block_height(12345);

// Use `set_*` methods to customize it later on.
ctx.set_block_timestamp(Timestamp::new(123));

// Use `as_*` methods to convert it to `{Mutable,Immutable,Sudo,Auth}Ctx`.
let mutable_ctx = ctx.as_mutable();
```


crates/types/Cargo.toml
```diff
@@ -33,6 +33,9 @@ sha2          = { workspace = true }
 sha3          = { workspace = true, optional = true }
 thiserror     = { workspace = true }
 
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+grug-crypto = { path = "../crypto" }
```
larry0x, line 37: `MockApi` uses grug-crypto, but it's not wasm-compatible so we need to exclude it from wasm builds.

crates/types/src/testing/querier.rs
```diff
@@ -0,0 +1,256 @@
...
+                    .app_configs
+                    .iter()
+                    .filter(|(k, _)| {
+                        if let Some(lower_bound) = &start_after {
+                            *k > lower_bound
+                        } else {
+                            true
+                        }
+                    })
+                    .take(limit.unwrap_or(u32::MAX) as usize)
```
larry0x, line 126: I don't like this... Better if we introduce a trait extension that adds a `may_take` method that takes a `Option<T>` where `T: Into<usize>`

PR #162, opened 2024-09-20
Simplify `Udec`

Rhaki:
Removed `const S: u32` as generic param for `Udec, hardcoded to `18`


Rhaki:
close in favor of https://github.com/left-curve/grug/pull/165

PR #164, opened 2024-09-21
Declarative check for bounds

larry0x:
A common use case is a contract needs to take a parameter that must be within a certain bound. E.g. a fee rate can be in in the range 0 <= fee rate < 1.

## Imperative approach

```rust
struct InstantiateMsg {
    pub fee_rate: Udec256,
}

fn instantiate(ctx: MutableCtx, msg: InstantiateMsg) -> anyhow::Result<Response> {
    ensure!(
        Udec256::ZERO <= fee_rate && fee_rate < Udec256::ONE,
        "fee rate is out of range!"
    );
}
```

## Declarative approach (this PR)

```rust
type FeeRate = Bounded<Uint256, FeeRateBounds>;

struct FeeRateBounds;

impl Bounds<Udec256> for FeeRateBounds {
    const MIN: Bound<Udec256> = Bound::Included<Udec256::ZERO>;
    const MAX: Bound<Udec256> = Bound::Excluded<Udec256::ONE>;
}

struct InstantiateMsg {
    pub fee_rate: FeeRate,
}

fn instantiate(ctx: MutableCtx, msg: InstantiateMsg) -> anyhow::Result<Response> {
    // no need to check for fee rate's value.
    // if `msg` successfully deserializes, fee rate must be within range.
}
```

The part of defining `FeeRate` and `FeeRateRange` can be abstracted using a macro:

```rust
define_bounds! {
    name = FeeRate,
    type = Udec256,
    min = Bound::Included(Udec256::ZERO),
    max = Bound::Excluded(Udec256::ONE)
}
```


larry0x:
This approach is limited by Rust's lack of support for:

1. Const generics of generic types, e.g.

   ```rust
   struct Bounded<T, const MIN: T, const MAX: T>(T);
   ```

   This doesn't work because the consts `MIN` and `MAX` can't have generic types.

2. Const generics can only be of number or char types. We can't have a `Uint128` const generic, for example.

crates/types/src/bound.rs
```diff
@@ -0,0 +1,78 @@
+use {
+    crate::{StdError, StdResult},
+    std::{marker::PhantomData, ops::Bound},
```
larry0x, line 3: What I'm not sure about this implementation is it uses `std::ops::Bound`, while we have another `grug_storage::Bound` of the same name. This can be confusing.

Should we move `grug_storage::Bound` here and use it instead?

larry0x: We're now using `grug_types::Bound`.

PR #167, opened 2024-09-23
M3 - Signed

Rhaki:
Generate Signed types.
Restore tests depending by signed


crates/math/Cargo.toml
```diff
@@ -14,6 +14,7 @@ bnum      = { workspace = true, features = ["borsh"] }
 borsh     = { workspace = true, features = ["derive", "de_strict_order"] }
 serde     = { workspace = true, features = ["derive"] }
 thiserror = { workspace = true }
+paste     = { workspace = true }
```
larry0x: Dependency isn't ordered alphabetically

crates/math/src/bytable.rs
```diff
@@ -61,49 +61,60 @@ macro_rules! impl_bytable_std {
     };
 }
 
+macro_rules! impl_bytable_singed_std {
```
larry0x: Typo: `signed` / `singed`

crates/math/src/bytable.rs
```diff
@@ -179,5 +223,74 @@ mod tests {
...
+
+            // Same thing for little endian
+            let recovered = Int128::from_le_bytes(number.to_le_bytes());
+            prop_assert_eq!(number, recovered);
+        }
+
+        /// Ensure the bytable methods work for `Int256`.
+        #[test]
+        fn integer_bytable_works_i256(number in any::<i128>()) {
+            let number = Int256::from(number);
```
larry0x: This test only tests integers within the i128 range. I'm going to modify it to test the full i256 range using the same approach as in `integer_bytable_works_u256`.

crates/math/src/udec.rs
```diff
@@ -681,19 +739,38 @@ generate_decimal! {
...
+    from_dec          = [],
+    doc               = "128-bit signed fixed-point number with 18 decimal places.",
+}
+
+generate_decimal! {
+    type              = Signed,
+    name              = Dec256,
+    inner_type        = I256,
+    inner_constructor = Int256::new_from_i128,
+    from_dec          = [],
```
larry0x, line 764: Should we add `From<Dec128>` here?

crates/math/src/uint.rs
```diff
@@ -38,19 +39,26 @@ where
     pub const fn number(&self) -> U {
         self.0
     }
+
+    pub const fn number_ref(&self) -> &U {
+        &self.0
+    }
```
larry0x, line 45: I think the more idiomatic way for defining these two functions is:

```rust
// takes a ref, return a ref
pub const fn number(&self) -> &U {
    &self.0
}

// takes an owned value, return an owned value
pub const fn into_number(self) -> U {
    self.0
}
```

Let's leave this to a later PR to fix.

crates/math/src/uint.rs
```diff
@@ -659,43 +678,69 @@ generate_uint! {
...
 }
 
-// -------------- additional constructor methods for Uint256/512 ---------------
+generate_uint! {
+    name = Int64,
+    inner_type = i64,
+    from_int = [],
+    from_std = [u32, u16, u8],
+    doc = "64-bit signed integer.",
+}
```
larry0x: Let's align the equation marks

crates/math/src/uint.rs
```diff
@@ -719,5 +764,36 @@ mod tests {
...
+        }
+        fn int512_const_constructor(input in any::<i128>()) {
+            let int512 = Int512::new_from_i128(input);
+            let output = int512.number().try_into().unwrap();
+            assert_eq!(input, output);
+        }
+    }
+
+    #[test]
+    fn singed_from_str() {
```
larry0x: Typo

PR #168, opened 2024-09-24
M4 - Traits reorganization

Rhaki:
Reorganize traits implementation for Math


crates/math/src/multiply_fraction.rs
```diff
@@ -13,3 +15,64 @@ where
...
+        }
+        if rhs.is_negative() {
+            return Err(MathError::negative_div(self, rhs));
+        }
+        if self.is_zero() {
+            return Ok(Self::ZERO);
+        }
+        self.checked_multiply_ratio_ceil(F::denominator(), rhs.numerator())
+    }
+}
```
Rhaki, line 78: Just noticed, are these negative check a thing?

larry0x: `Uint` was intended for only positive numbers, so we have it. Now I think it can be removed

PR #175, opened 2024-09-27
M6 - Int tests

Rhaki:
Note: beside tests there is a change in trait `Signed`.
The function `abs` has been renamed into `checked_abs` and now return `MathResult<Self>`, which allow to handle the Overflow of signed number at `MIN`.


crates/math/src/int.rs
```diff
@@ -362,53 +362,278 @@ impl Int512 {
...
-        assert_eq!(-Int512::new_from_i128(100), Int512::new(I512::from(-100)))
-    }
+                let serialized_vec = serde_json::to_vec(&original).unwrap();
+                assert_eq!(serialized_vec, format!("\"{}\"", sample).as_bytes());
+
+                let parsed: Int::<_> = serde_json::from_str(&serialized_str).unwrap();
+                assert_eq!(parsed, original);
+
+                let parsed: Int::<_> = serde_json::from_slice(&serialized_vec).unwrap();
+                assert_eq!(parsed, original);
```
larry0x, line 521: Instead of `serde_json`, use the `Json{Ser,De}Ext` traits here

PR #178, opened 2024-10-01
M7 - Dec tests

Rhaki:


Rhaki:
Close. See https://github.com/left-curve/left-curve/pull/182

PR #179, opened 2024-10-02
Improve `MultiplyRatio`

Rhaki:
- added `checked_multiply_ratio` that has the same logic of native rust integer;
- added tests.


Rhaki:
Replaced by https://github.com/left-curve/left-curve/pull/183

PR #182, opened 2024-10-03
M7 - Dec tests

Rhaki:


grug/math/src/decimal.rs
```diff
@@ -11,27 +11,168 @@ pub trait Decimal: Sized {
 impl<U> Decimal for Dec<U>
 where
     Self: FixedPoint<U>,
-    U: Number + Copy + PartialEq,
+    U: Number + Copy + PartialEq + Sign + NumberConst,
```
larry0x, line 14: Let's order the trait bounds such that all number traits come first, then all the std traits

grug/math/src/decimal.rs
```diff
@@ -11,27 +11,168 @@ pub trait Decimal: Sized {
...
-        // Method 2 is faster because Rem is roughly as fast as or slightly
-        // faster than Div, while Sub is significantly faster than Mul.
-        //
-        // This flooring operation in fact can never fail, because flooring an
-        // unsigned decimal goes down to 0 at most. However, flooring a _signed_
-        // decimal may underflow.
-        Ok(Self(self.0 - self.0.checked_rem(Self::DECIMAL_FRACTION)?))
+        let rem = self.0.checked_rem(Self::DECIMAL_FRACTION)?;
+
+        match (rem == Int::ZERO, rem.is_negative()) {
```
larry0x, line 19: Instead of `rem == Int::ZERO` maybe `rem.is_zero()`?

grug/math/src/number.rs
```diff
@@ -125,11 +101,15 @@ where
...
-            .map(Self)
+        let clos = || {
+            self.0
+                .checked_full_mul(*other.numerator())?
+                .checked_div(Self::DECIMAL_FRACTION.into_next())?
+                .checked_into_prev()
+                .map(Self)
+        };
+
+        clos().map_err(|_| MathError::overflow_mul(self, other))
```
larry0x, line 112: It's unnecessary to define the `clos`? You can simply do:

```rust
self.0
    .checked_full_mul(*other.numerator())?
    .checked_div(Self::DECIMAL_FRACTION.into_next())?
    .checked_into_prev()
    .map(Self)
    .map_err(|_| MathError::overflow_mul(self, other))
```

grug/math/src/number.rs
```diff
@@ -141,24 +121,28 @@ where
...
+                    exp = (exp - 1) / 2;
+                }
             }
-        }
 
-        self.checked_mul(y)
+            self.checked_mul(y)
+        };
+
+        clos().map_err(|_| MathError::overflow_pow(self, exp))
```
larry0x, line 145: You don't need the closure here either. Just do:

```rust
{
    // ...
}
.map_err(|_| MathError::overflow_pow(self, exp))
```

grug/math/src/number.rs
```diff
@@ -913,113 +881,1033 @@ mod tests {
...
-    int_test!( wrapping_add
-        method = |_0| {
-            let max = bt(_0, Int::MAX);
-            assert_eq!(max.wrapping_add(Int::ONE), Int::MIN);
+    int_test!( saturating_add
+        inputs = {
+            u128 = {
+                passing: [
+                    (Int::MAX - Int::ONE, Int::ONE, Int::MAX),
+                    (Int::MAX, Int::ONE, Int::MAX)
```
larry0x, line 889: Let's add trailing commas

grug/math/src/test_utils.rs
```diff
@@ -189,3 +192,226 @@ macro_rules! int_test {
...
+    /// `Dec::raw(Int::<U>::ONE)`
+    const LEFTOVER: Self;
+}
+
+impl<U> Leftover for Dec<U>
+where
+    Int<U>: NumberConst,
+{
+    const LEFTOVER: Dec<U> = Dec::raw(Int::<U>::ONE);
+}
```
larry0x, line 386: I think this can be added to the `FixedPoint` trait?

Also "left over" isn't a good name. Maybe "step" or "tick"?

grug/math/src/test_utils.rs
```diff
@@ -189,3 +192,226 @@ macro_rules! int_test {
...
+//         dec256 = {
+//             passing: [
+//             ],
+//             failing: [
+//             ]
+//         }
+//     }
+//     method = |_0d: Dec<_>, passing, failing| {
+//     }
+// );
```
larry0x, line 417: These can be deleted?

grug/math/src/number.rs
Rhaki: It's because if the mul overflow, another error another error is raised. In this way any error raised is converted into OverflowPow

grug/math/src/test_utils.rs
Rhaki: I've added in test utils because idk if this has some needed outside test purpose

grug/math/src/test_utils.rs
Rhaki: Yeah. I forgot it (used as model to be copy pasted)

grug/math/src/number.rs
larry0x: I see, you're right. Still you can do:

```rust
{
    // ...
}
.map_err(|_| MathError::overflow_mul(self, other))
```

PR #196, opened 2024-10-09
refractor: `Binary` & `ByteArray`

Rhaki:
Currently, the `Binary` structure is simply a wrapper for `Vec<u8>` that serializes/deserializes the bytes to/from a `base64-encoded string`. The `ByteArray `structure performs the same function. This proposal aims to unify both types into a single structure called `B64`, with a generic `T `that can represent either a `Vec<u8> `or a fixed-size array `[u8; N]`. The resulting types will be:

```rust
type Binary = B64<Vec<u8>>
type ByteArray<const N: usize> = B64<[u8; N]>
```


Rhaki:
refractor implemebted by #198

PR #221, opened 2024-10-14
refactor(grug): revamp generic error

larry0x:
- Define `GenericResult<T>` as an alias to `Result<T, String>`
- Replace `StdError::Generic` with `StdError::Host`
- Add `AppError::Guest`. The error message includes the contract address, the wasm export function being called, and the error message returned by the contract.


larry0x:
I'm going to do a bigger revamp of errors.

larry0x:
The bigger revamp appears challenging, so we go with this.

PR #225, opened 2024-10-16
fix(dango): passkey authentication

j0nl1:
We've updated the passkey signature behavior. Previously, the contract was responsible for generating the client_data. However, this approach posed potential limitations, as the clientDataJson could expand over time. To address this, we now allow the UI to provide the client_data and ensure the challenge matches the transaction sign bytes.


dango/types/src/auth.rs
```diff
@@ -55,13 +56,14 @@ pub struct Metadata {
 #[grug::derive(Serde)]
 pub struct PasskeySignature {
     pub authenticator_data: Binary,
-    pub origin: String,
-    pub cross_origin: bool,
+    pub client_data: Binary,
```
larry0x, line 59: Can you use `grug::Json` for `client_data` here? It would be more readable.

dango/types/src/auth.rs
```diff
@@ -55,13 +56,14 @@ pub struct Metadata {
...
     pub sig: ByteArray<64>,
 }
 
-/// Passkey client data.
-#[grug::derive(Serde)]
+/// Passkey client data
+/// We use serde instead of grug because it does not support unknown fields.
```
larry0x, line 64: Let's just remove `deny_unknown_fields` from grug derive. I'm going to push a commit

dango/auth/src/lib.rs
```diff
@@ -91,13 +89,14 @@ pub fn authenticate_tx(
...
-    .to_json_vec()?
-    .hash256();
+    let sign_bytes = ctx.api.sha2_256(
+        &SignDoc {
+            messages: tx.msgs,
+            chain_id: ctx.chain_id,
+            sequence: metadata.sequence,
+        }
+        .to_json_vec()?,
+    );
```
larry0x, line 99: Use the host function for hashing instead. By removing the sha256 logic, contract sizes go down by ~10 KB.

PR #228, opened 2024-10-16
refractor(grug): use `Option<&str>` instead of generic for label

Rhaki:
…tion<Into<String>> for label


larry0x:
Do you think we should do the same for `Client::instantiate`?

Rhaki:
Yeah

larry0x:
> Yeah

Can you make the change in this pr

PR #233, opened 2024-10-17
refactor(grug): use singleton tuple variants for `Query`

larry0x:
Just like #232, but for query


Rhaki:
There is only one conceptual "problem" now. The singleton tuple variants of enum `Query` now implement the `QueryRequest` trait. Iin the `TestSuite` and `QuerierWrapper`, the functions `query_wasm_smart`accept as a parameter `R: QueryRequest`. So at compile time we allow to use this function and pass as argument a singleton variant of the `Query` enum, which could lead to confusion.

larry0x:
Removed `QueryRequest` implementations.

Rhaki (approved):

Issue #235, opened 2024-10-19
dango: remove token creation fee


larry0x:
It adds friction to apps and doesn't really prevent spamming (you can use IBC to spam.)

PR #243, opened 2024-10-20
feat(dango): remove denom creation fee from token factory

larry0x:
Closes: #235

I'm not sure if we want to remove this yet. Need to think...


larry0x:
Replaced by #244

PR #248, opened 2024-10-21
refractor(grug): `AppCtx`

Rhaki:


grug/app/src/vm.rs
```diff
@@ -266,27 +218,25 @@ where
...
+    let code = CODES.load(&ctx.storage, code_hash)?;
 
     // Create the providers
-    let querier = QuerierProvider::new(vm.clone(), storage.clone(), gas_tracker.clone(), block);
-    let storage = StorageProvider::new(storage, &[CONTRACT_NAMESPACE, &contract]);
+    let querier = QuerierProvider::new(ctx.clone());
+    let storage = StorageProvider::new(ctx.storage.clone(), &[CONTRACT_NAMESPACE, &contract]);
 
-    Ok(vm.build_instance(
+    Ok(ctx.vm().clone().build_instance(
```
larry0x: No need to clone the VM here. You need a `&mut VM`, just take the `ctx` parameter as mut.

grug/app/src/ctx.rs
```diff
@@ -0,0 +1,110 @@
...
+impl<VM> AppCtx<VM>
+where
+    VM: Vm,
+{
+    pub fn vm(&self) -> &VM {
+        &self.vm
+    }
+
+    /// Downcast the `AppCtx<VM>` to `AppCtx<Undefined<()>>`
+    pub fn downcast(&self) -> AppCtx<Undefined<()>> {
```
larry0x: No need to clone here. The method can take a `self` and return an `AppCtx<Undefined>`.

grug/app/src/app.rs
```diff
@@ -1,5 +1,7 @@
+use crate::AppCtx;
 #[cfg(feature = "abci")]
 use grug_types::{JsonDeExt, JsonSerExt};
+
```
larry0x: Make sure the format the `use` statements correctly... sometimes the formatter isn't smart enough to do it

grug/app/src/app.rs
```diff
@@ -294,43 +294,37 @@ where
...
 
-        match do_authenticate(
-            self.vm.clone(),
-            Box::new(buffer),
-            gas_tracker.clone(),
-            block,
-            &tx,
-            AuthMode::Check,
-        ) {
+        ctx.gas_tracker = GasTracker::new_limited(tx.gas_limit);
```
larry0x: I don't like replacing a field in the context. It feels unsafe, not a good pattern.

grug/vm-wasm/tests/tester.rs
```diff
@@ -101,7 +101,7 @@ fn immutable_state() {
     suite
         .send_message_with_gas(
             &mut accounts["sender"],
-            1_000_000,
+            2_000_000,
```
larry0x, line 104: Now that we have user pay also for gas consumed in `withhold_fee`, we need to increase this for the test to pass.

grug/vm-wasm/tests/transfer.rs
```diff
@@ -88,16 +88,12 @@ fn transfers_with_insufficient_gas_limit() {
...
-    // a host function call (in which case, a `VmError::OutOfGas`). We can only
-    // say that the error has to be one of the two. Therefore, we simply ensure
-    // the error message contains the word "gas".
+    // For this test to work, the tx should request enough gas to pass `withhold_fee`,
+    // but not enough to cover the actual transfer.
+    // In experience, 200k gas works.
     let outcome = suite.send_message_with_gas(
         &mut accounts["sender"],
-        100_000,
+        200_000,
```
larry0x, line 96: Same here, need to increase the cost (see comment above)

grug/app/src/app.rs
```diff
@@ -719,36 +672,26 @@ where
     VM: Vm + Clone,
     AppError: From<VM::Error>,
 {
+    let gas_tracker = ctx.replace_gas_tracker(GasTracker::new_limitless());
```
larry0x, line 675: While we let users pay for `withhold_fee`, we can't let user pay for `finalize_fee`.

Because taxman uses the gas consumption of everything up to before `finalize_fee` to calculate how much fee to charge.

PR #249, opened 2024-10-21
fix: small build fixes

apollo-sturdy:
Ran into some issues when trying to run devnet and frontend. Fixes applied:

- Add prerequisite to have Node.js v20.18.0. I was on 16 or 18 (can't recall) and it only worked once upgraded to 20.18.0. Perhaps some lesser version works, but not sure.
- Allow CORS from any origin in devnet. Otherwise frontend can't connect to RPC node in docker container.
- Add docker compose file for easy launch of devnet with `docker compose up` and `docker compose rm` to reset.

### DIFF
--- a/justfile
+++ b/justfile
@@ -79,8 +79,12 @@ docker-build-devnet:
 
 # Start a devnet from genesis
 start-devnet:
-  docker run -it -p 26657:26657 -p 26656:26656 {{DEVNET_NAME}}:{{DEVNET_VERSION}}
+  docker run --name devnet -it -p 26657:26657 -p 26656:26656 {{DEVNET_NAME}}:{{DEVNET_VERSION}}
 
 # Restart a devnet that have been previous stopped
-restart-devnet container_id:
-  docker start -i $1
+restart-devnet:
+  docker start -i devnet
+
+# Remove a devnet
+remove-devnet:
+  docker rm -f devnet
--- a/justfile
+++ b/justfile
@@ -79,12 +79,12 @@ docker-build-devnet:
 
 # Start a devnet from genesis
 start-devnet:
-  docker run --name devnet -it -p 26657:26657 -p 26656:26656 {{DEVNET_NAME}}:{{DEVNET_VERSION}}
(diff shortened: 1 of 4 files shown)


j0nl1:
The minimum required version is `18.0.0`. You can find the requirement in the [package.json](https://github.com/left-curve/left-curve/blob/main/package.json#L41) or the [SDK Readme](https://github.com/left-curve/left-curve/blob/main/sdk/README.md?plain=1#L62)

Would you mind to update the version and maybe point to sdk readme?
We also could include a nvmrc file to automatically change the version.

Regarding the docker compose, there are already commands defined in just to launch the devnet container and restart it.
What we could do instead of providing a container-id, we could run the container with a specific name so it would be easier to re-start.

apollo-sturdy:
> The minimum required version is `18.0.0`. You can find the requirement in the [package.json](https://github.com/left-curve/left-curve/blob/main/package.json#L41) or the [SDK Readme](https://github.com/left-curve/left-curve/blob/main/sdk/README.md?plain=1#L62)
> 
> Would you mind to update the version and maybe point to sdk readme? We also could include a nvmrc file to automatically change the version.
> 
> Regarding the docker compose, there are already commands defined in [just](https://github.com/left-curve/left-curve/blob/main/justfile#L81) to launch the devnet container and restart it. What we could do instead of providing a container-id, we could run the container with a specific name so it would be easier to re-start.

Thanks for the input. Changed the docs and updated the justfile to use named docker container and remove the compose file.

justfile
```diff
@@ -79,8 +79,12 @@ docker-build-devnet:
 
 # Start a devnet from genesis
 start-devnet:
-  docker run -it -p 26657:26657 -p 26656:26656 {{DEVNET_NAME}}:{{DEVNET_VERSION}}
+  docker run --name devnet -it -p 26657:26657 -p 26656:26656 {{DEVNET_NAME}}:{{DEVNET_VERSION}}
```
larry0x, line 82: Maybe use the chain ID as container name?

apollo-sturdy: Sure. will fix

PR #253, opened 2024-10-22
chore(sdk): align query types with grug and actions

j0nl1:


larry0x:
I've updated the new `CodeResponse` which was introduced in #255

Issue #263, opened 2024-10-23
bug(dango): ensure only the safe account itself can vote in a proposal


larry0x:
In `authenticate`, we should ensure the sender is the Safe account itself (`ctx.sender == ctx.contract`)

larry0x:
completed by #409

PR #267, opened 2024-10-24
fix(dango): safe account auth issues

### LINKED ISSUE #262: bug: voter authentication in Safe account
Given a Safe account, consider such a sequence of events:

1. the Safe starts with 3 members: alice, bob, charlie
2. prop 1 is created to remove alice from the member set
3. prop 2 is created
4. prop 1 is passed and executed; at this time, prop 2 is still voting
5. alice attempts to vote in prop 2

Since prop 2 was created when alice was still a member, she should be able to vote.

However, in `dango_auth::authenticate`, the contract determines whether alice can vote by checking whether alice's username is associated with the Safe account, which isn't, because they are already disassociated in step 4.

Rhaki:
Closed #262 #263


larry0x:
replaced by #409

Issue #265, opened 2024-10-23
grug: add `#[must_use]` to functions returning Outcome or TxOutcome


larry0x:
This allows clippy to warn if you don't unwrap the outcome with `.should_succeed()`

larry0x:
completed by #270

PR #270, opened 2024-10-24
feat(grug): Outcome `must-use`

Rhaki:
solve #265.
instead to use `#[must_use]` on every function that return an Outcome, the attribute `#[must_use]` is added directly to the various `Outcome` structures

### DIFF
--- a/grug/testing/src/suite.rs
+++ b/grug/testing/src/suite.rs
@@ -16,6 +16,7 @@ use {
 
 // ------------------------------- UploadOutcome -------------------------------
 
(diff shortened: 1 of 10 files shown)


grug/testing/src/suite.rs
```diff
@@ -16,6 +16,7 @@ use {
 
 // ------------------------------- UploadOutcome -------------------------------
 
+#[must_use = "UploadOutcome must be checked for success or error with `should_succeed`, `should_fail` or similar methods."]
```
larry0x, line 19: Let's always use the oxford comma: https://english.stackexchange.com/a/233072

PR #274, opened 2024-10-25
feat: extensions

Rhaki:


larry0x:
I think you need to rebuild the contracts in vm-wasm/testdata

PR #277, opened 2024-10-27
feat(dango): oracle contract

larry0x:
the way this oracle contract works:

- it is to be defined as a cronjob. let's call the time interval it's execute the **epoch** (say, every 30 seconds)
- during an epoch, a set of **guardians** submit prices
- at the end of each epoch, the oracle contract computes the median price of each denom

a few remaining issues:

- add exponential moving average (EMA)
- how to deal with AMM LP tokens? compute them onchain or have guardians submit the prices as well?
- chain owner should be able to add/remove guardians
- handle the case of possible chain halt - oracle should exclude feeds that are too old

### DIFF
--- a/dango/oracle/src/execute.rs
+++ b/dango/oracle/src/execute.rs
@@ -50,6 +50,7 @@ pub fn cron_execute(ctx: SudoCtx) -> StdResult<Response> {
     // Load all feeds submitted by guardians since the last cron execute.
     for res in PRICE_FEEDS.prefix_range(ctx.storage, None, None, Order::Ascending) {
         let ((denom, _guardian), feed) = res?;
+
         // We expect there to be ~20 guardians, so pre-allocate a vector
         // with capacity of 20.
         denom_feeds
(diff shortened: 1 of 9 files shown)


larry0x:
an alternative solution is to use pyth. simply submit pyth data signed by wormhole guardian set.

let me research how much effort is needed for that.

dango/oracle/src/execute.rs
```diff
@@ -0,0 +1,102 @@
...
+}
+
+#[cfg_attr(not(feature = "library"), grug::export)]
+pub fn cron_execute(ctx: SudoCtx) -> StdResult<Response> {
+    let cfg = CONFIG.load(ctx.storage)?;
+
+    let mut denom_feeds = BTreeMap::<Denom, Vec<Udec128>>::new();
+
+    // Load all feeds submitted by guardians since the last cron execute.
+    for res in PRICE_FEEDS.prefix_range(ctx.storage, None, None, Order::Ascending) {
```
apollo-sturdy, line 51: Why `prefix_range` instead of `range`?

dango/oracle/src/execute.rs
```diff
@@ -0,0 +1,102 @@
...
+        if num_feeds < cfg.quorum as usize {
+            continue;
+        }
+
+        // The price feeds must be sorted in order to find the median.
+        feeds.sort();
+
+        // Find the median price.
+        // Here as an optimization, we use bitwise operators (`&` and `>>`)
+        // instead of division or modulo.
```
apollo-sturdy, line 76: Could add these to a trait to have better clarity of code? E.g. `fn is_even`

larry0x:
Replaced by #284

PR #279, opened 2024-10-28
feature(grug): typed `GasTracker`

Rhaki:
`GasTracker` now includes a generic parameter, allowing it to be defined as either `Limited`, `Limitless`, or `Undefined` in specific functions.


larry0x:
closing this as we agree it's too complex of a solution to a small problem

PR #280, opened 2024-10-28
feat(grug): app config buffer

Rhaki:


Rhaki:
close in favor of #304

PR #282, opened 2024-10-29
feat(grug): ABCI++ prepare proposal

larry0x:
adds support for state-less, synchronous proposal preparer.

in the future, we can consider supporting stateful or async ones.


grug/testing/src/suite.rs
```diff
@@ -243,16 +289,6 @@ where
...
 
-        // Sanity check: the number of tx results returned by the app should
-        // equal the number of txs.
-        assert_eq!(
-            num_txs,
-            block_outcome.tx_outcomes.len(),
-            "sent {} txs but received {} tx results; something is wrong",
-            num_txs,
-            block_outcome.tx_outcomes.len()
-        );
```
larry0x, line 254: this sanity check is no longer correct, as now PP can insert new txs into the list

PR #291, opened 2024-10-30
fix(grug): extra node being saved in JMT subtree creation

bugarela:
Hello :octocat: 

This includes a test from a scenario I found with [Quint](https://quint-lang.org) where, after some combination of operations, we end up with a leaf node being saved both on bit `0` and `010`.

### DIFF
--- a/grug/jellyfish-merkle/src/tree.rs
+++ b/grug/jellyfish-merkle/src/tree.rs
@@ -1060,7 +1060,7 @@ mod tests {
         // Before doing any pruning, check nodes and orphans are correct.
         assert_tree(
             &storage,
-            vec![
+            [
                 (0, ROOT_BITS),
                 (0, BitArray::from_bits(&[0])),
                 (0, BitArray::from_bits(&[1])),
@@ -1078,7 +1078,7 @@ mod tests {
                 (3, ROOT_BITS),
                 // v5 tree is empty
             ],
-            vec![
+            [
                 (1, 0, ROOT_BITS),
                 (1, 0, BitArray::from_bits(&[0])),
                 (1, 0, BitArray::from_bits(&[0, 1])),
@@ -1101,7 +1101,7 @@ mod tests {
         TREE.prune(&mut storage, 1).unwrap();
         assert_tree(
             &storage,
-            vec![
+            [
                 (0, BitArray::from_bits(&[1])),
                 (0, BitArray::from_bits(&[0, 1, 0])),
                 (1, ROOT_BITS),
@@ -1112,7 +1112,7 @@ mod tests {
                 (2, BitArray::from_bits(&[0])),
                 (3, ROOT_BITS),
             ],
-            vec![
+            [
                 (2, 0, BitArray::from_bits(&[0, 1, 0])),
                 (2, 1, ROOT_BITS),
                 (2, 1, BitArray::from_bits(&[0])),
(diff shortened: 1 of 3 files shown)


larry0x:
Thanks @bugarela. I've applied the fix. Please see this commit: https://github.com/left-curve/left-curve/pull/291/commits/bd229e4511c9dcd8e5feb69354b5efd29587ba6c

grug/jellyfish-merkle/src/tree.rs
```diff
@@ -375,45 +375,67 @@ impl<'a> MerkleTree<'a> {
...
+        match (batch.len(), existing_leaf) {
+            // The subtree to be created is empty: do nothing.
+            (0, None) => Ok(Outcome::Unchanged(None)),
+            // The subtree to be created contains exactly one node, which is an
+            // existing leaf node.
+            (0, Some(leaf_node)) => Ok(Outcome::Unchanged(Some(Node::Leaf(leaf_node)))),
+            // The subtree to be created contains exactly one node, which is a
+            // new leaf node.
+            // This case requires special attention: we don't save the node yet,
+            // because it the path may be collapsed if it's sibling gets deleted.
```
bugarela, line 387:
```suggestion
            // because the path may be collapsed if it's sibling gets deleted.
```

grug/jellyfish-merkle/src/tree.rs
```diff
@@ -375,45 +375,67 @@ impl<'a> MerkleTree<'a> {
...
+                Ok(Outcome::Updated(Node::Leaf(LeafNode {
                     key_hash,
                     value_hash,
-                })
+                })))
             },
+            // The subtree to be created contains more 2 or more nodes.
+            // Recursively create the tree. Return the subtree's root, an
+            // internal node.
+            // Note that in this scenario, we certain don't need to collapse the
```
bugarela, line 398:
```suggestion
            // Note that in this scenario, we certainly don't need to collapse the
```

bugarela (commented):
> smaller instances of this problem that would make for a simpler test case

Can you tell me the smaller instance? I would like to keep both using use the `#[test_case]` macro

grug/jellyfish-merkle/src/tree.rs
larry0x: There's another typo in this sentence (its/it's)

bugarela: updated the comment

bugarela:
I think you edited my question comment instead of replying to it :sweat_smile: 
The smallest I could find this far is also two consecutive applies, but with less ops:
```bluespec
Set(
  { key_hash: [0, 1, 1, 0], op: Insert([8]) },
  { key_hash: [1, 1, 0, 0], op: Insert([2]) }
),
Set(
  { key_hash: [0, 1, 1, 0], op: Delete },
  { key_hash: [1, 1, 0, 0], op: Delete },
  { key_hash: [1, 0, 0, 0], op: Insert([2]) }
)
```

This results in the same leaf at `_` (root) and at `1` in version 2:
```bluespec
{ key_hash: [], version: 1 } ->
  Internal({
    left_child: Some({ hash: Map([0, 0] -> Raw([1, 0, 1, 1, 0, 8]), [0] -> Hash), version: 1 }),
    right_child: Some({ hash: Map([0, 0] -> Raw([1, 1, 1, 0, 0, 2]), [0] -> Hash), version: 1 })
  }),
{ key_hash: [], version: 2 } -> Leaf({ key_hash: [1, 0, 0, 0], value_hash: [2] }),
{ key_hash: [0], version: 1 } -> Leaf({ key_hash: [0, 1, 1, 0], value_hash: [8] }),
{ key_hash: [1], version: 1 } -> Leaf({ key_hash: [1, 1, 0, 0], value_hash: [2] }),
{ key_hash: [1], version: 2 } -> Leaf({ key_hash: [1, 0, 0, 0], value_hash: [2] })
```

PR #301, opened 2024-11-04
feat: Implement health check for margin account borrowing

### LINKED ISSUE #299: dango: health factor
The dango-lending contract stores a **collateral power** for each collateral asset, and a `Market` for each borrowable asset:

```rust
const COLLATERAL_POWERS: Map<&Denom, Udec128> = Map::new("collateral_power");

const MARKETS: Map<&Denom, Market> = Map::new("market");
```

- An asset may be a collateral asset but not a borrowable asset, e.g. wstETH, stATOM, LP tokens. But typically all borrowable assets are also collateral assets, such that when a margin account borrows an asset, this asset counts both as collateral and debt.
- Collateral powers are to be bounded in the range `[0, 1)`. An asset with lower volatility and more abundant liquidity gets bigger collateral power, vise versa.
- We may store all collateral powers in a single `Item<BTreeMap<Denom, Udec128>>` if we don't expect to support too many collateral assets.

Suppose:
- a margin account has collateral assets $A_1, A_2, \dots, A_n$ and debts $B_1, B_2, \dots, B_m$
- the price of an asset $X$ is $P_X$
- the collateral power of an asset $X$ is $C_X$

The account's **utilization** is:

```math
U = \frac{ \sum_{j=1}^m P_{B_j} B_j }{ \sum_{i=1}^n C_{A_i} P_{A_i} A_i }
```

In the `backrun` function, the margin account asserts $U \leq 1$. If not true, it throws an error to revert the transaction.

The frontend should additionally have a `max_ltv`, somewhat smaller than 1, such as 95%. It should warn or prevent users from doing anything that results in their utilization going bigger than this, such that their account isn't instantly liquidated.

apollo-sturdy:
Closes #299


dango/account-margin/Cargo.toml
```diff
@@ -18,9 +18,10 @@ crate-type = ["cdylib", "rlib"]
 library = []
 
 [dependencies]
-anyhow      = { workspace = true }
-dango-auth  = { workspace = true }
-dango-types = { workspace = true }
-grug        = { workspace = true }
+anyhow        = { workspace = true }
+dango-auth    = { workspace = true }
+dango-lending = { workspace = true }
```
larry0x, line 23: this should be imported with `features = ["library"]`?

dango/lending/src/execute.rs
```diff
@@ -41,13 +50,52 @@ fn update_markets(
...
+) -> anyhow::Result<Response> {
+    // Ensure only chain owner can update collateral powers.
+    ensure!(
+        ctx.sender == ctx.querier.query_config()?.owner,
+        "Only the owner can update collateral powers"
+    );
+
+    let mut collateral_powers = COLLATERAL_POWERS.load(ctx.storage)?;
+    collateral_powers.insert(denom, power);
+    COLLATERAL_POWERS.save(ctx.storage, &collateral_powers)?;
```
larry0x, line 80: use `COLLATERAL_POWERS.may_update`?

dango/lending/src/execute.rs
```diff
@@ -41,13 +50,52 @@ fn update_markets(
...
+pub fn delist_collateral(ctx: MutableCtx, denom: Denom) -> anyhow::Result<Response> {
+    // Ensure only chain owner can delist collateral.
+    ensure!(
+        ctx.sender == ctx.querier.query_config()?.owner,
+        "Only the owner can delist collateral tokens"
+    );
+
+    let mut collateral_powers = COLLATERAL_POWERS.load(ctx.storage)?;
+    collateral_powers.remove(&denom);
+    COLLATERAL_POWERS.save(ctx.storage, &collateral_powers)?;
```
larry0x, line 94: use `COLLATERAL_POWERS.may_update`?

dango/lending/src/query.rs
```diff
@@ -64,3 +77,64 @@ pub fn query_debts(
...
+    querier: &QuerierWrapper,
+    margin_account: Addr,
+    debts: Coins,
+    collateral_powers: BTreeMap<Denom, CollateralPower>,
+) -> anyhow::Result<HealthResponse> {
+    // Calculate the total value of the debts.
+    let mut total_debt_value = Udec128::ZERO;
+    for coin in debts {
+        let price = query_oracle_price(querier, &coin.denom)?;
+        total_debt_value += coin.amount.checked_into_dec()? * price;
```
Rhaki, line 106: use dedicated function to mul `Int` with `Dec`
```suggestion
        total_debt_value += coin.amount.checked_mul_dec(price)?;
```

dango/lending/src/query.rs
```diff
@@ -64,3 +77,64 @@ pub fn query_debts(
...
+    }
```
Rhaki, line 107: using `try_fold` is a better practice for this situation

dango/lending/src/query.rs
```diff
@@ -64,3 +77,64 @@ pub fn query_debts(
...
+
+    // Calculate the total value of the account's collateral adjusted for the collateral power.
+    let mut total_adjusted_collateral_value = Udec128::ZERO;
+    for (denom, power) in collateral_powers {
+        let price = query_oracle_price(querier, &denom)?;
+        let collateral_balance = querier.query_balance(margin_account, denom)?;
+        let collateral_value = collateral_balance.checked_into_dec()? * price;
+        total_adjusted_collateral_value += collateral_value * power.into_inner();
+    }
```
Rhaki, line 116: There is an efficiency problem in this cycle. If the lending allows multiple collaterals, it is necessary to perform a query_balance and a query_price for each collateral. With time and the listing of new collaterals this part becomes very expensive. A solution would be to allow in the entry_point receive only the tokens managed by the lending pool and to save the various denominations. Then in calculate_account_health, perform a query on the balance of the tokens at the bank, if the quantity is 0 remove from the map managed by entry_point receive and go to the next, otherwise query the price and add the value to total_adjusted_collateral_value

dango/lending/src/query.rs
apollo-sturdy: Not sure if its better practice really? Just different style no?

I actually did that first, but felt that it was harder to read.

Happy to change it to `try_fold` if both you and @larry0x prefer it.

dango/lending/src/execute.rs
apollo-sturdy: Do you prefer may_update and remove setting the Item to an empty BTreeMap in instantiate entrypoint? Or fine to use update and assume it always exists?

dango/lending/src/query.rs
apollo-sturdy: But that gives an Int? Is there an equivalent that gives Dec?

dango/lending/src/query.rs
apollo-sturdy: Hmm. @larry0x How many collateral tokens do we expect? Also how much in general do we care about such performance issues? I figured we're not in such a constrained environment as EVM and can easily handle iterating over a few dozen tokens without issue?

I also think it's a bit of a shame to restrict margin accounts only to hold those tokens which are listed as collateral tokens. Personally don't like making such restriction unless really necesssary. The more freedom people have in the ways they want to use the product the better imo.

@Rhaki you bring up a good point though that we don't need to query the price from the oracle if the collateral balance is zero. So will change the order here and continue in the loop if balance is zero.

dango/lending/src/execute.rs
```diff
@@ -150,3 +201,43 @@ pub fn borrow(ctx: MutableCtx, coins: Coins) -> anyhow::Result<Response> {
...
+            .query_wasm_raw(account_factory, ACCOUNTS.path(ctx.sender))?
+            .ok_or_else(|| anyhow!(
+                "account {} is not registered in account factory",
+                ctx.sender
+            ))?
+            .deserialize_borsh::<Account>()?
+            .params
+            .is_margin(),
+        "Only margin accounts can borrow and repay"
+    );
```
larry0x, line 221: Is this necessary? If an account is not a margin account, it shouldn't have debt in the first place.

dango/lending/src/execute.rs
```diff
@@ -150,3 +201,43 @@ pub fn borrow(ctx: MutableCtx, coins: Coins) -> anyhow::Result<Response> {
...
+    // Update the sender's liabilities
+    DEBTS.may_update(ctx.storage, ctx.sender, |maybe_debts| {
+        let mut debts = maybe_debts.unwrap_or_default();
+
+        debts
+            .deduct_many(ctx.funds)
+            .map_err(|_| anyhow!("Cannot repay more than the debts"))?;
+
+        Ok(debts)
+    })?;
```
larry0x, line 240: I think we should just refund any excess payment. This is because debts are compounded every block. It may be 100 USDC this block but 100.05 USDC next block. To deal with this the common approach is to send a little excess (e.g. 101 USDC) and refund the unused.

larry0x:
The file `dango/types/src/lending.rs` is getting too big. You consider splitting it into multiple files as we do for `dango_types::account_factory`.

dango/types/src/lending.rs
```diff
@@ -62,4 +108,10 @@ pub enum QueryMsg {
         start_after: Option<Addr>,
         limit: Option<u32>,
     },
+    /// Returns all collateral powers.
+    #[returns(BTreeMap<Denom, CollateralPower>)]
+    CollateralPowers {},
+    /// Queries the health of a margin account.
+    #[returns(HealthResponse)]
+    Health { account: Addr },
```
larry0x, line 116: I think the `Health` query should be in the margin account instead of lending. Lending only concerns about lending out assets and compound interests. The margin account concerns health and liquidation.

Collateral powers probably should be an app config.

dango/types/src/oracle/price.rs
```diff
@@ -8,9 +8,15 @@ pub type PrecisionedPrice = Price<Defined<u8>>;
 
 #[grug::derive(Serde, Borsh)]
 pub struct Price<P = Defined<u8>> {
+    /// The price of the token in its humanized form. I.e. the price of 1 ATOM, rather than 1 uatom.
```
larry0x, line 11: As a general style guideline, comments are limited to 80 characters long.

dango/types/src/oracle/price.rs
```diff
@@ -26,9 +32,20 @@ impl PrecisionlessPrice {
...
+    /// is 10^6 uatom, so the precision is 6.
     pub fn precision(&self) -> u8 {
         self.precision.into_inner()
     }
+
+    /// Returns the value of a given unit amount. E.g. if this Price represents
+    /// the price in USD of one ATOM, then this function will return the value in USD of
+    /// the given number of uatom.
+    pub fn value_of_unit_amount(&self, unit_amount: Uint128) -> Udec128 {
+        self.humanized_price * Udec128::new(*unit_amount.inner())
```
larry0x, line 46: Instead of `Udec128::new(*unit_amount.inner())` use `Udec128::raw(unit_amount)`?

dango/types/src/oracle/price_source.rs
```diff
@@ -19,4 +27,22 @@ impl PriceSource {
...
+    pub fn raw_query_price(
+        &self,
+        querier: &QuerierWrapper,
+        oracle: Addr,
+    ) -> anyhow::Result<PrecisionedPrice> {
+        match self {
+            Self::Pyth { id, precision } => {
+                let price = querier
+                    .query_wasm_raw(oracle, PRICES.path(*id))?
+                    .ok_or(anyhow::anyhow!("Price not found for denom: {}", id))?
```
larry0x, line 41: Import `anyhow::anyhow` and invoke it as `anyhow!` here instead of `anyhow::anyhow!`

grug/storage/src/codec.rs
```diff
@@ -17,6 +17,7 @@ pub trait Codec<T> {
 // ----------------------------------- borsh -----------------------------------
 
 /// Represents the Borsh encoding scheme.
+#[derive(Clone)]
 pub struct Borsh;
```
larry0x, line 21: Might as well derive all common traits (debug, clone, copy, partialeq, eq) for all other `Codec`'s as well

dango/lending/src/query.rs
```diff
@@ -64,3 +77,62 @@ pub fn query_debts(
...
+
+        // As an optimization, don't query the price if the collateral balance is zero.
+        if collateral_balance.is_zero() {
+            continue;
+        }
+
+        let price = dango_oracle::raw_query_price(querier, &denom)?;
+        let collateral_value = price.value_of_unit_amount(collateral_balance);
+        total_adjusted_collateral_value += collateral_value * power.into_inner();
+    }
```
larry0x, line 114: if there's an asset that's both a collateral and a debt, then we'd be repeat the oracle query twice

dango/lending/src/query.rs
```diff
@@ -64,3 +77,62 @@ pub fn query_debts(
...
+
+    if total_adjusted_collateral_value.is_zero() {
+        bail!("The account has no collateral");
+    }
+
+    // Calculate the utilization rate.
+    let utilization_rate = total_debt_value / total_adjusted_collateral_value;
```
larry0x, line 121: for this you can just do `.checked_div`

dango/lending/src/query.rs
larry0x: also every time you do `dango_oracle::raw_query_price`, you do `query_app_config` again, which is a waste...

dango/oracle/src/query.rs
```diff
@@ -43,6 +48,25 @@ fn query_price(ctx: ImmutableCtx, denom: Denom) -> anyhow::Result<PrecisionedPri
...
+        .query_wasm_raw(oracle, PRICE_SOURCES.path(denom))?
+        .ok_or(anyhow::anyhow!(
+            "Price source not found for denom: {}",
+            denom
+        ))?
+        .deserialize_borsh::<PriceSource>()?
+        .raw_query_price(querier, oracle)?;
+
+    Ok(price)
+}
```
larry0x, line 68: i don't like this pattern: logics scattered across multiple contracts. if a query is initiated from contract X, then the query logic should be in contract X.

dango/types/src/lending.rs
```diff
@@ -10,27 +13,56 @@ pub static NAMESPACE: LazyLock<Part> = LazyLock::new(|| Part::new_unchecked("len
...
 pub enum ExecuteMsg {
     /// Apply updates to markets.
     UpdateMarkets(BTreeMap<Denom, MarketUpdates>),
+    /// Set the collateral power for a denom.
+    SetCollateralPower {
+        denom: Denom,
+        power: CollateralPower,
+    },
+    /// Delist a collateral token. Removes it from the collateral power map.
+    DelistCollateral { denom: Denom },
```
larry0x, line 65: i think we should put collateral powers in the app config. in the case we don't need to implement these two methods

dango/lending/src/execute.rs
apollo-sturdy: True. Can remove it.

dango/lending/src/execute.rs
apollo-sturdy: Ah yeah makes sense. I can change it.

dango/lending/src/query.rs
apollo-sturdy: > if there's an asset that's both a collateral and a debt, then we'd be repeat the oracle query twice

Yeah that's what I meant by creating a querier that caches results. Thought you said it was not necessary. Can implement that if you want?

dango/oracle/src/query.rs
apollo-sturdy: Ok. I just figured probably more contracts will need the same query and then unless we want to duplicate this code everywhere we need one place to put it and I figured in the oracle itself made the most sense.

But if we make the "OracleQuerier" mentioned above then it can live there and be a separate file in the types package?

dango/types/src/lending.rs
apollo-sturdy: Sure. Can move it

dango/types/src/oracle/price.rs
apollo-sturdy: But that's not the same? `Udec128::new(42)` gives `"42.0"` whereas `Udec128::raw(10)` gives `0.00000000000000042`?

dango/account-margin/src/execute.rs
```diff
@@ -22,7 +31,41 @@ pub fn instantiate(ctx: MutableCtx, _msg: InstantiateMsg) -> anyhow::Result<Resp
...
+        .ok_or_else(|| anyhow!("collateral powers not found"))?
+        .deserialize_borsh::<BTreeMap<Denom, CollateralPower>>()?;
+
+    // Calculate the utilization rate.
+    let health = calculate_account_health(&ctx.querier, ctx.contract, debts, collateral_powers)?;
+
+    // If the utilization rate is greater than 1, the account is undercollateralized.
+    ensure!(
+        health.utilization_rate <= Udec128::ONE,
+        "The action would make the account undercollateralized. Utilization rate after action: {}. Total debt: {}. Total collateral: {}",
```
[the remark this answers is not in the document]
apollo-sturdy: What about this case where it's multiple sentences? Makes sense to have punctation between or?

dango/lending/src/query.rs
apollo-sturdy: Do we prefer to use checked math ops always? Any pros and cons to consider here? I'm actually not sure why checked ops are recommended in CosmWasm. Cause actually non-checked versions give a much better backtrace with line number.

dango/types/src/lending.rs
apollo-sturdy: What about Markets? What are the pros and cons with app config?

dango/types/src/oracle/price.rs
```diff
@@ -26,9 +34,20 @@ impl PrecisionlessPrice {
...
     pub fn precision(&self) -> u8 {
         self.precision.into_inner()
     }
+
+    /// Returns the value of a given unit amount. E.g. if this Price represents
+    /// the price in USD of one ATOM, then this function will return the value
+    /// in USD of the given number of uatom.
+    pub fn value_of_unit_amount(&self, unit_amount: Uint128) -> Udec128 {
+        self.humanized_price * Udec128::new(*unit_amount.inner())
+            / Udec128::new(10u128.pow(self.precision.into_inner() as u32))
```
Rhaki, line 49: This could lead overflow. For tokens with large precision like eth (1eth = 10^18 wei) self.humanized_price * Udec128::new(*unit_amount.inner()) overflow with even 1 eth at the current price (~3k usd). The correct implementation is:
```suggestion
    pub fn value_of_unit_amount(&self, unit_amount: Uint128) -> StdResult<Udec128> {
        self.humanized_price * Udec128::checked_from_ratio(unit_amount, 10u128.pow(self.precision.into_inner()))
    }
```

dango/account-margin/src/query.rs
```diff
@@ -1,16 +1,90 @@
...
+/// Calculates the health of a margin account.
+pub fn calculate_account_health(
+    querier: &QuerierWrapper,
+    margin_account: Addr,
+    debts: Coins,
+    collateral_powers: BTreeMap<Denom, CollateralPower>,
+) -> anyhow::Result<HealthResponse> {
+    // Calculate the total value of the debts.
+    let mut total_debt_value = Udec128::ZERO;
+    for coin in debts {
```
Rhaki, line 39: The function `dango_oracle::raw_query_price(querier, &coin.denom)?` query the oracle contract. If the user has different debts is a waste. Add ad an optional param `maybe_oracle: Option<Addr>`. If it's None, query for it, otherwise use the one passed

dango/account-margin/src/query.rs
```diff
@@ -1,16 +1,90 @@
...
+    let mut total_adjusted_collateral_value = Udec128::ZERO;
+    for (denom, power) in collateral_powers {
+        let collateral_balance = querier.query_balance(margin_account, denom.clone())?;
+
+        // As an optimization, don't query the price if the collateral balance is zero.
+        if collateral_balance.is_zero() {
+            continue;
+        }
+
+        let price = dango_oracle::raw_query_price(querier, &denom)?;
```
Rhaki, line 54: Same problem here for querying the oracle_addr multiple times

dango/types/src/oracle/price.rs
apollo-sturdy: Good catch. Thanks!

dango/lending/src/execute.rs
```diff
@@ -150,3 +151,30 @@ pub fn borrow(ctx: MutableCtx, coins: Coins) -> anyhow::Result<Response> {
...
+
+    // Update the sender's liabilities
+    DEBTS.may_update(ctx.storage, ctx.sender, |maybe_debts| {
+        let mut debts = maybe_debts.unwrap_or_default();
+
+        // Deduct the sent coins from the account's debts, saturating at zero.
+        let remainders = debts.saturating_deduct_many(ctx.funds)?;
+
+        // Refund the remainders to the sender
+        msgs.push(Message::transfer(ctx.sender, remainders)?);
```
larry0x, line 174: This should only be done if remainder is non-empty

dango/lending/src/execute.rs
```diff
@@ -150,3 +151,30 @@ pub fn borrow(ctx: MutableCtx, coins: Coins) -> anyhow::Result<Response> {
...
+pub fn repay(ctx: MutableCtx) -> anyhow::Result<Response> {
+    // Ensure all sent coins are whitelisted
+    for coin in &ctx.funds {
+        ensure!(
+            MARKETS.has(ctx.storage, coin.denom),
+            "Invalid denom. Only whitelisted denoms can be repaid."
+        );
+    }
+
+    let mut msgs = vec![];
```
larry0x, line 164: Instead of a `Vec<Message>`, you can create a

```rust
let mut maybe_msg = None;
```

and then

```rust
Ok(Response::new().may_add_message(maybe_msg))
```

dango/account-safe/src/execute.rs
```diff
@@ -277,6 +278,10 @@ mod tests {
                     // Address below don't matter for this test.
                     ibc_transfer: Addr::mock(0),
                     oracle: Addr::mock(1),
+                    lending: Addr::mock(0), // doesn't matter for this test
+                },
+                lending: LendingAppConfig {
+                    collateral_powers: btree_map! {},
                 },
```
larry0x: For these we can implement `Default` trait on `AppConfig`, then simply do:

```rust
AppConfig {
    addresses: AppAddresses {
        account_factory: ACCOUNT_FACTORY,
        ..Default::default()
    },
    ..Default::default()
}
```

dango/types/src/lending.rs
```diff
@@ -10,6 +10,46 @@ pub static NAMESPACE: LazyLock<Part> = LazyLock::new(|| Part::new_unchecked("len
...
+#[grug::derive(Serde)]
+pub struct CollateralPowerBounds;
+
+impl Bounds<Udec128> for CollateralPowerBounds {
+    const MAX: Option<Bound<Udec128>> = Some(Bound::Inclusive(Udec128::ONE));
+    const MIN: Option<Bound<Udec128>> = Some(Bound::Exclusive(Udec128::ZERO));
+}
+
+/// A decimal bounded by the collateral power bounds.
+pub type CollateralPower = Bounded<Udec128, CollateralPowerBounds>;
```
larry0x: As commented earlier, collateral power is not the concern of the lending contract, so it shouldn't be in this file. I'm going to move this to the `account/margin.rs`.

dango/types/src/oracle/price.rs
```diff
@@ -55,3 +82,28 @@ impl TryFrom<PriceFeed> for PrecisionlessPrice {
         })
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use grug::NumberConst;
+
+    use super::*;
```
larry0x: Combine the `use` statements into a single one:

```rust
use {grug::NumberConst, super::*};
```

and add the separator comment:

```rust
// ----------------------------------- tests -----------------------------------
```

dango/types/src/oracle/price.rs
```diff
@@ -55,3 +82,28 @@ impl TryFrom<PriceFeed> for PrecisionlessPrice {
...
+            humanized_ema: Udec128::ONE,
+            timestamp: 0,
+            precision: Defined::new(18),
+        };
+
+        // Value of 100M ETH at $100M = $100000T
+        let value = price
+            .value_of_unit_amount(Uint128::new(100_000_000u128 * 10u128.pow(18)))
+            .unwrap();
+        println!("{value:?}");
```
larry0x: This print statement can be removed?

grug/types/src/coins.rs
```diff
@@ -207,6 +207,44 @@ impl Coins {
...
+        self.deduct(coin.clone())?;
+
+        Coin::new(coin.denom, 0)
+    }
+
+    /// Deduct all coins from another `Coins`, saturating at zero. Returns a
+    pub fn saturating_deduct_many(&mut self, coins: Self) -> StdResult<Self> {
+        let mut remainders = Self::new();
+
+        for coin in coins.into_iter() {
```
larry0x: `.into_iter()` is unnecessary. You can just do

```rust
for coin in coins {
    // ...
}
```

grug/types/src/coins.rs
```diff
@@ -516,4 +554,45 @@ mod tests {
...
+
+    #[test]
+    fn saturating_deduct_many() {
+        // Deduct less than available
+        let mut coins = mock_coins();
+        let deduct = Coins::try_from([
+            ("uatom", Uint128::new(100)),
+            ("umars", Uint128::new(100)),
+            ("uosmo", Uint128::new(789)),
+        ])
```
larry0x: For tests, instead of `Coins::try_from` we can simply do `Coins::new_unchecked`.

grug/storage/src/codec.rs
```diff
@@ -5,10 +5,11 @@ use {
     },
     prost::Message,
     serde::{de::DeserializeOwned, ser::Serialize},
+    std::fmt::Debug,
 };
 
 /// A marker that designates encoding/decoding schemes.
-pub trait Codec<T> {
+pub trait Codec<T>: Debug + Clone + Copy + PartialEq + Eq {
```
larry0x: What's the purpose of this? I deleted it and everything still compiles and works...

dango/lending/src/execute.rs
```diff
@@ -150,3 +151,30 @@ pub fn borrow(ctx: MutableCtx, coins: Coins) -> anyhow::Result<Response> {
...
 }
+
+pub fn repay(ctx: MutableCtx) -> anyhow::Result<Response> {
+    // Ensure all sent coins are whitelisted
+    for coin in &ctx.funds {
+        ensure!(
+            MARKETS.has(ctx.storage, coin.denom),
+            "Invalid denom. Only whitelisted denoms can be repaid."
+        );
+    }
```
larry0x, line 162: Note: in practice, we don't need to assert this, because once we have implemented interest rate, we need to update interest rate state here, so if the denom isn't whitelisted, this is will naturally fail.

dango/account-margin/src/query.rs
```diff
@@ -1,16 +1,86 @@
...
+    debts: Coins,
+    collateral_powers: BTreeMap<Denom, CollateralPower>,
+) -> anyhow::Result<HealthResponse> {
+    let app_cfg: AppConfig = querier.query_app_config()?;
+
+    // Calculate the total value of the debts.
+    let total_debt_value = debts.into_iter().try_fold(Udec128::ZERO, |acc, coin| {
+        let price = querier.query_price(app_cfg.addresses.oracle, &coin.denom)?;
+        Ok::<_, anyhow::Error>(acc + price.value_of_unit_amount(coin.amount)?)
+    })?;
```
larry0x, line 43: I think this code is just so much worse than just using a `for`-loop.

Especially the `Ok::<_, anyhow::Error>` part

dango/account-margin/src/query.rs
```diff
@@ -1,16 +1,86 @@
...
     NEXT_SEQUENCE.current(storage)
 }
+
+/// Calculates the health of a margin account.
+pub fn calculate_account_health(
+    querier: &QuerierWrapper,
+    margin_account: Addr,
+    debts: Coins,
+    collateral_powers: BTreeMap<Denom, CollateralPower>,
+) -> anyhow::Result<HealthResponse> {
```
larry0x, line 36: Two optimizations can be done here:

1. Cache prices: if a denom is both a debt and a collateral, we shouldn't query the price twice.
2. If the account doesn't have any debt, then it's healthy for sure. No need to compute the adjusted collateral value.

Let's do these in a later PR.

dango/account-margin/src/query.rs
```diff
@@ -1,16 +1,86 @@
...
+        }
+
+        let price = querier.query_price(app_cfg.addresses.oracle, &denom)?;
+        let collateral_value = price.value_of_unit_amount(collateral_balance)?;
+        total_adjusted_collateral_value += collateral_value * power.into_inner();
+    }
+
+    if total_adjusted_collateral_value.is_zero() {
+        bail!("the account has no collateral");
+    }
```
larry0x, line 62: If an account has no collateral but also no debt, it's considered healthy. In this case should return a utilization rate of zero instead of throwing error here.

larry0x: This:

```rust
let utilization_rate = if total_debt_value.is_zero() {
    Udec128::ZERO
} else if total_adjusted_collateral_value.is_zero() {
    Udec128::MAX
} else {
    total_debt_value / total_adjusted_collateral_value
};
```

dango/lending/Cargo.toml
```diff
@@ -21,6 +21,7 @@ library = []
 anyhow                = { workspace = true }
 dango-account-factory = { workspace = true, features = ["library"] }
 dango-auth            = { workspace = true }
+dango-oracle          = { workspace = true, features = ["library"] }
```
larry0x, line 24: This dependency seems unnecessary

apollo-sturdy: Thanks. Perhaps worth adding cargo-machete to the just file.

dango/account-margin/src/query.rs
apollo-sturdy: Opened PRs #331 and #332 for this

dango/account-margin/src/query.rs
apollo-sturdy: That's what I thought, but @Rhaki had strong opinions that this was better, so I changed it.

dango/lending/src/execute.rs
apollo-sturdy: True, but that error would probably be quite confusing I'd assume?

grug/storage/src/codec.rs
apollo-sturdy: I figured since we wanted all Codecs to implement these we might as well require it so that we don't forget if we add new Codecs.

dango/types/src/oracle/price.rs
apollo-sturdy: Yes, was just for testing, forgot to remove.

dango/types/src/lending.rs
```diff
@@ -10,6 +10,46 @@ pub static NAMESPACE: LazyLock<Part> = LazyLock::new(|| Part::new_unchecked("len
...
+// -------------------------------- Bounds -------------------------------------
+
+/// Defines the bounds for a loan-to-value ratio: 0 < LoanToValue < 1.
+#[grug::derive(Serde)]
+pub struct LoanToValueBounds;
+
+impl Bounds<Udec128> for LoanToValueBounds {
+    const MAX: Option<Bound<Udec128>> = Some(Bound::Exclusive(Udec128::ONE));
+    const MIN: Option<Bound<Udec128>> = Some(Bound::Exclusive(Udec128::ZERO));
+}
```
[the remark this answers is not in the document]
apollo-sturdy: Oops, no. Not used anymore. Just missed to remove them.

dango/account-safe/src/execute.rs
apollo-sturdy: Mm yes I thought the same. Suggested it in Discord

PR #302, opened 2024-11-04
docs(grug): JMT invariants

bugarela:
Hello :octocat: 

This is work in progress on invariants for the JMT manipulation. We are defining these in a markdown document so they are nicer to read and can include nicely rendered English descriptions as well as the Quint definitions. The Quint code get's tangled (I'm using [lmt](https://github.com/driusan/lmt)) to a Quint file that we check, so this is a live document - we can automate these checks in the CI at some point.

I recommend reviewing the rendered markdown, as the raw version contains snippets inside commented code that are necessary for the tangled code to work, but are not really in scope at this point. The rendered document should show only core data structures and the invariants themselves.

We are opening this draft PR to potentially receive feedback on the invariants we were able to define so far.

### DIFF
--- a/grug/jellyfish-merkle/spec/invariants.md
+++ b/grug/jellyfish-merkle/spec/invariants.md
@@ -169,7 +169,7 @@ This inserts a line break that is not rendered in the markdown
 
 In a tree, all nodes except from the root should have a parent. There should be no dangling nodes. If we find a node with key 1001, there should be a node for 100. At some other iteration, we'll also check that 100 has a parent (that is, 10), and so on.
 
-*Status:* FALSE due to https://github.com/left-curve/left-curve/pull/291
+*Status:* TRUE
 
 ```bluespec apply_state_machine.qnt +=
   /// Make sure the tree encoded in the map forms a tree (everyone has a parent, except for the root)
@@ -229,7 +229,7 @@ This inserts a line break that is not rendered in the markdown
 
 Leafs should never be in the middle of a tree. If a there is a node with a key hash that is a prefix of the key hash of a leaf, then this is not a proper tree.
 
-*Status:* FALSE due to https://github.com/left-curve/left-curve/pull/291
+*Status:* TRUE
 
 ```bluespec apply_state_machine.qnt +=
   /// Make sure that the map encodes a tree. In particular, there is no internal node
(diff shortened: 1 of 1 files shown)


larry0x:
Thanks @bugarela, looks good! I don't have any other invariants to add.

> Internal nodes share the version with at least one child

This on indeed isn't an invariant. You're correct about the case of deletions. Another case is where a batch is empty. In this case, we get a new root node at the new version, but the root node's two subtrees are unchanged.

> Status: FALSE due to https://github.com/left-curve/left-curve/pull/291

As #291 is merged, these invariants should be TRUE now?

bugarela:
> As #291 is merged, these invariants should be TRUE now?

yes @larry0x, they hold on the simulations I ran so far :smile: I'll update this document with the new status.

grug/jellyfish-merkle/spec/invariants.md
```diff
@@ -0,0 +1,583 @@
...
+<!--
+This inserts a line break that is not rendered in the markdown
+```bluespec apply_state_machine.qnt +=
+
+```
+-->
+
+### Internal nodes share the version with at least one child
+
+Given the explanation around `versionInv` from above, we had the (wrong) intuition, that since updates always push their version up the tree, every internal node should have the version of at least one of it children. However, the intuition is misleading in the case of a delete, where a parent gets a new version, but there may not be a node at the spot where the deleted nodes had been. However, for reference, we keep the formula here as it might be useful for understanding in the future.
```
josef-widder, line 359:
```suggestion
Given the explanation around `versionInv` from above, we had the (wrong) intuition, that since updates always push their version up the tree, every internal node should have the version of at least one of it children. However, the intuition is misleading 
- in the case of a delete, where a parent gets a new version, but there may not be a node at the spot where
    the deleted nodes had been. However, for reference, we keep the formula here as it might be useful for 
    understanding in the future.
- in the case of adding an empty batch, where we get a new root node at the new version, but the root 
    node's the two subtrees are unchanged.
```

josef-widder (commented):
I just added an explanation on the false `denseVersionsInv`

grug/jellyfish-merkle/spec/invariants.md
bugarela: Updating your version to move the `However...` part below
```suggestion
Given the explanation around `versionInv` from above, we had the (wrong) intuition, that since updates always push their version up the tree, every internal node should have the version of at least one of it children. However, the intuition is misleading 
- in the case of a delete, where a parent gets a new version, but there may not be a node at the spot where the deleted nodes had been. 
- in the case of applying an empty batch, where we get a new root node at the new version, but the root node's subtrees are unchanged.
However, for reference, we keep the formula here as it might be useful for understanding in the future.
```

bugarela:
Closing in favor of #375 which includes this changes.

PR #304, opened 2024-11-05
feat(grug): cfg-buffer

Rhaki:


larry0x:
Let's close this for now. We just need to make sure if we change taxman to a different address, then `finalize_fee` must not fail.

PR #310, opened 2024-11-06
refractor(grug): `Tx` generics

### LINKED ISSUE #309: grug: generic metadata and credential in txs
Currently we use `Json` for metadata and credential in `Tx`:

```rust
struct Tx {
    // ...
    data: Json,
    credential: Json,
}
```

The drawback of this is when the account contract does `authenticate`, it needs to deserialize the data and credential:

```rust
let credential = tx.credential.deserialize_json::<dango_types::auth::Credential>()?;
```

Perhaps we can use generics instead:

```rust
struct Tx<D = Json, C = Json> {
    // ...
    data: D,
    credential: C,
}
```

In dango-types, we can create a dango-specific tx type:

```rust
use crate::auth::{Credential, Metadata};

type Tx = grug::Tx<Metadata, Credential>;
```

Rhaki:
Closes #309

### DIFF
--- a/grug/types/src/tx.rs
+++ b/grug/types/src/tx.rs
@@ -23,12 +23,12 @@ pub type Label = LengthBounded<String, 1, 128>;
 
 /// A transaction.
 #[derive(Serialize, Deserialize, BorshSerialize, BorshDeserialize, Debug, Clone, PartialEq, Eq)]
-pub struct Tx {
+pub struct Tx<D = Json, C = Json> {
     pub sender: Addr,
     pub gas_limit: u64,
     pub msgs: Vec<Message>,
-    pub data: Json,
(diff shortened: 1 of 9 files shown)


grug/types/src/tx.rs
```diff
@@ -23,12 +23,12 @@ pub type Label = LengthBounded<String, 1, 128>;
 
 /// A transaction.
 #[derive(Serialize, Deserialize, BorshSerialize, BorshDeserialize, Debug, Clone, PartialEq, Eq)]
-pub struct Tx {
+pub struct Tx<D = Json, C = Json> {
```
larry0x, line 26: I feel it may be better to ensure C and D are `DeserializeOwned`

PR #314, opened 2024-11-07
feat(grug): Indexer initialization

penso:
Add native indexer ability to Grug and Dango.

- Add `indexer/*` crates
- Add `Indexer` trait in Grug
- Add non-blocking indexer used for Dango, can be enabled running `start --indexer-enabled`
- Change testing suite, using `NullIndexer` by default (doesn't do anything) to keep existing tests as it is
- Add tests with non_blocking indexer enabled
- Add SQL indexer using `sea-orm` with sqlite and postgres
- Add SQL entities and migrations
- Add justfile
- Store blocks, transactions, messages, events into SQL tables


penso:
@larry0x We still need to figure out what happens when the process gets killed or `Ctrl-c` to avoid missing indexing. We should manage signals.

Is the code called by `do_commit` atomic? Can't that committing code itself be killed at a wrong time, while storing a block so half transactions have been processed?

larry0x:
> Is the code called by `do_commit` atomic? Can't that committing code itself be killed at a wrong time, while storing a block so half transactions have been processed?

We use a `WriteBatch` to write data to RocksDB in a single batch. It should be atomic in my understanding

grug/app/Cargo.toml
```diff
@@ -10,9 +10,10 @@ license       = { workspace = true }
 categories    = { workspace = true }
 
 [features]
-default = []
+default = ["indexer"]
```
penso, line 13: Good way to do this?

Cargo.toml
```diff
@@ -123,6 +124,12 @@ dango-testing         = { path = "dango/testing" }
 dango-token-factory   = { path = "dango/token-factory" }
 dango-types           = { path = "dango/types" }
 
+# Indexer packages
+indexer-core          = { path = "indexer/core" }
+indexer-sql           = { path = "indexer/sql" }
+indexer-sql-migration = { path = "indexer/sql-migration" }
+assertor              = { version = "0.0", git = "https://github.com/google/assertor" }
```
penso, line 131: See the testing extension already existing.

dango/cli/Cargo.toml
```diff
@@ -13,6 +13,10 @@ categories    = { workspace = true }
 name = "dango"
 path = "src/main.rs"
 
+[features]
+default = ["indexer"]
```
penso, line 17: Should we use this for proposer?

dango/cli/src/start.rs
```diff
@@ -23,13 +26,48 @@ pub struct StartCmd {
...
         let db = DiskDb::open(data_dir)?;
         let vm = WasmVm::new(self.wasm_cache_capacity);
+        #[cfg(feature = "indexer")]
+        let app = {
+            // I tried using `null_indexer` but since indexer is a App generic parameter, the
+            // created `App` has a different type and I'd have to duplicate lots of code, use an
+            // `enum Indexer` with all potential indexer, or use dyn IndexerTrait. Instead I added
+            // a `indexing_enabled` field on `App` to not call the indexer so you can disable the
+            // indexer when running this binary.
+            let mut indexer = non_blocking_indexer::Indexer::async_new_with_database_url(
```
penso, line 50: Could use a method using a generic.

grug/app/src/app.rs
```diff
@@ -46,9 +47,26 @@ pub struct App<DB, VM, PP = NaiveProposalPreparer> {
     /// Related config in CosmWasm:
     /// <https://github.com/CosmWasm/wasmd/blob/v0.51.0/x/wasm/types/types.go#L322-L323>
     query_gas_limit: u64,
+    #[cfg(feature = "indexer")]
+    pub indexer_app: INDEXER,
+    #[cfg(feature = "indexer")]
+    pub indexing_enabled: bool,
```
penso, line 53: Without the feature enabled this code won't compile since the generic isn't used anywhere.

indexer/sql/src/non_blocking_indexer.rs
```diff
@@ -0,0 +1,350 @@
...
+    fn call<F>(&self, closure: F) -> error::Result<()>
+    where
+        F: Future<Output = Result<(), indexer_core::error::Error>> + Send + 'static,
+    {
+        match self.runtime.as_ref() {
+            Some(runtime) => {
+                runtime.block_on(closure)?;
+            },
+            None => {
+                tokio::task::block_in_place(|| self.handle.block_on(closure))?;
```
penso, line 194: Look at potential issue when saving DB transaction

larry0x:
Suggest the following changes:

1. Remove the `indexer` feature from grug-app.

   `App` will always have an indexer. If we don't want to do any indexing, just use `NullIndexer`.

2. Rename the `INDEXER` generic in `App` to `ID`.

3. Reorder the generics in `App` as follows:

   ```rust
   pub struct App<
       DB,
       VM,
       PP = NaiveProposalPreparer,
       VE = NullVoteExtender,
       ID = NullIndexer,
   > where
       DB: Db,
       VM: Vm,
       PP: ProposalPreparer,
       VE: VoteExtender,
       ID: Indexer,
   { /* ... */ }
   ```

   (The `VE` generic and `VoteExtender` trait don't exist yet; they will be added in a later PR.)

4. The `indexer_enabled` flag shouldn't be in `App`. It should be part of the indexer's state, i.e.

   ```rust
   struct Indexer {
       pub enabled: bool,
   }
   ```

5. Remove the word "Trait" from `IndexerTrait` and other traits that are named this way.

   If there's a collision in naming (e.g. we have a struct called `Indexer` and we want to implement the `Indexer` trait for it), do it like this:

   ```rust
   struct Indexer { /* ... */ }
   
   impl grug_app::Indexer for Indexer { /* ... */ }
   ```

6. Move `Indexer` trait to grug-app and place it in the `trait.rs` file. `NullIndexer` should also be in grug-app, in an `indexer.rs` file. (similar to how we place `NaiveProposalPreparer` in `grug/app/src/proposal.rs`)

   The `trait.rs` file may be too long this way; in a later PR we can split it into smaller files.

   The idea is that everything in the `grug/` directory should be at the bottom of the dependency tree. It shouldn't depend on anything in the `indexer/` directory.

grug/app/src/app.rs
```diff
@@ -148,10 +153,10 @@ where
     pub fn do_prepare_proposal(&self, txs: Vec<Bytes>, max_tx_bytes: usize) -> Vec<Bytes> {
         let txs = self
             ._do_prepare_proposal(txs.clone(), max_tx_bytes)
-            .unwrap_or_else(|err| {
+            .unwrap_or_else(|_err| {
                 #[cfg(feature = "tracing")]
                 error!(
-                    err = err.to_string(),
+                    err = _err.to_string(),
```
Rhaki, line 159: Don't add `_` on `_err` if we use it

grug/app/src/app.rs
```diff
@@ -48,21 +51,23 @@ pub struct App<DB, VM, PP = NaiveProposalPreparer> {
...
 }
 
-impl<DB, VM, PP> App<DB, VM, PP>
+impl<DB, VM, PP, ID> App<DB, VM, PP, ID>
 where
     DB: Db,
     VM: Vm + Clone,
+    ID: Indexer,
     PP: ProposalPreparer,
     AppError: From<DB::Error> + From<VM::Error> + From<PP::Error>,
```
Rhaki, line 72: Similar to DB, VM, PP, add the following bound: `AppError: From<ID::Error>`. This allow to use `?` instead of `unwrap` for `Indexer` functions.

grug/app/src/app.rs
```diff
@@ -282,14 +291,20 @@ where
...
                 block,
-                tx,
+                tx.clone(),
                 AuthMode::Finalize,
-            ));
+            );
+
+            self.indexer
+                .index_transaction(&block, &tx, &tx_outcome)
+                .unwrap();
```
Rhaki, line 305: Use `?` instead of `unwrap`,adding bound `AppError: From<ID::Error>`

grug/app/src/app.rs
```diff
@@ -318,11 +333,15 @@ where
...
 
-        Ok(BlockOutcome {
+        let block_outcome = BlockOutcome {
             app_hash: app_hash.unwrap(),
             cron_outcomes,
             tx_outcomes,
-        })
+        };
+
+        self.indexer.index_block(&block, &block_outcome).unwrap();
```
Rhaki, line 342: Use `?` instead of `unwrap`,adding bound `AppError: From<ID::Error>`

grug/app/src/app.rs
```diff
@@ -331,6 +350,10 @@ where
         #[cfg(feature = "tracing")]
         tracing::info!(height = self.db.latest_version(), "Committed state");
 
+        if let Some(block_height) = self.db.latest_version() {
+            self.indexer.post_indexing(block_height).unwrap();
```
Rhaki, line 354: Use `?` instead of `unwrap`,adding bound `AppError: From<ID::Error>`

grug/app/src/indexer.rs
```diff
@@ -0,0 +1,57 @@
...
+    }
+}
+
+impl Default for NullIndexer {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl super::Indexer for NullIndexer {
```
Rhaki, line 22: if possible for our types, import the type witout specify the path.
```suggestion
impl Indexer for NullIndexer {
```

grug/testing/Cargo.toml
```diff
@@ -24,6 +24,19 @@ rand               = { workspace = true }
 serde              = { workspace = true }
 tracing            = { workspace = true }
 tracing-subscriber = { workspace = true }
+indexer-sql        = { workspace = true, features = ["tracing"] }
+assertor           = { workspace = true }
```
Rhaki, line 28: Sort dependencies alphabetically

grug/testing/Cargo.toml
```diff
@@ -24,6 +24,19 @@ rand               = { workspace = true }
...
+
+[dependencies.sea-orm]
+version = "1.1.0"
```
Rhaki, line 31: Usually we add all of our dep in workspace Cargo.toml

grug/testing/tests/indexing.rs
```diff
@@ -0,0 +1,149 @@
...
+        .set_owner("owner")
+        .build();
+
+    let to = accounts["owner"].address;
+
+    let _outcome = suite.send_message_with_gas(
+        &mut accounts["sender"],
+        2000,
+        Message::transfer(to, Coins::one(denom.clone(), 2_000).unwrap()).unwrap(),
+    );
```
Rhaki, line 30: Use `should_success` from `ResultExt` trait instead to catch in a var
```suggestion
   suite.send_message_with_gas(
        &mut accounts["sender"],
        2000,
        Message::transfer(to, Coins::one(denom.clone(), 2_000).unwrap()).unwrap(),
    ).should_success();
```

grug/testing/tests/indexing.rs
```diff
@@ -0,0 +1,149 @@
...
+
+            Ok::<(), sea_orm::DbErr>(())
+        })
+        .expect("Can't commit txn");
+
+    let _outcome = suite.send_message_with_gas(
+        &mut accounts["sender"],
+        3000,
+        Message::transfer(to, Coins::new()).unwrap(),
+    );
```
Rhaki, line 69: use `should_succeed` instead

grug/testing/tests/indexing.rs
```diff
@@ -0,0 +1,149 @@
...
+                .order_by_desc(entity::blocks::Column::BlockHeight)
+                .one(&indexer.context.db)
+                .await
+                .expect("Can't fetch blocks");
+            assert_that!(block).is_some();
+            assert_that!(block.unwrap().block_height).is_equal_to(2);
+            Ok::<(), sea_orm::DbErr>(())
+        })
+        .expect("Can't commit txn");
+}
```
Rhaki, line 85: Since you are handling the errors with expect in the Future, you can return () and not a Result.
expect at line 84 can then removed (impossible to reach)

grug/testing/tests/indexing.rs
```diff
@@ -0,0 +1,149 @@
...
+        .set_owner("owner")
+        .build();
+
+    let to = accounts["owner"].address;
+
+    let _outcome = suite.send_message_with_gas(
+        &mut accounts["sender"],
+        2000,
+        Message::transfer(to, Coins::one(denom.clone(), 2_000).unwrap()).unwrap(),
+    );
```
Rhaki, line 110: use `should_succeed` instead

grug/testing/tests/indexing.rs
```diff
@@ -0,0 +1,149 @@
...
+
+            let events = entity::events::Entity::find()
+                .all(&indexer.context.db)
+                .await
+                .expect("Can't fetch events");
+            assert_that!(events).is_not_empty();
+
+            Ok::<(), sea_orm::DbErr>(())
+        })
+        .expect("Can't commit txn");
```
Rhaki, line 148: Since you are handling the errors with expect in the Future, you can return () and not a Result.
expect at line 148 can then removed (impossible to reach)

indexer/sql-migration/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "indexer-sql-migration"
+version = "0.1.0"
+edition = "2021"
+publish = false
```
Rhaki, line 5: Allign equal symbols

Rhaki, line 5:
```suggestion
name    = "indexer-sql-migration"
version = "0.1.0"
edition = "2021"
publish = false
```

indexer/sql-migration/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+
+[lib]
+name = "migration"
+path = "src/lib.rs"
+
+[dependencies]
+async-std = { version = "1", features = ["attributes", "tokio1"] }
+
+[dependencies.sea-orm-migration]
+version = "1.1.0"
```
Rhaki, line 15: use workspace

indexer/sql/Cargo.toml
```diff
@@ -0,0 +1,42 @@
...
+[features]
+tracing = ["dep:tracing"]
+
+[dependencies]
+thiserror             = { workspace = true }
+tokio                 = { workspace = true }
+tracing               = { workspace = true, optional = true }
+anyhow                = { workspace = true }
+indexer-sql-migration = { workspace = true }
+uuid                  = { version = "1", features = ["serde", "v4"] }
```
Rhaki, line 24: use workspace

indexer/sql/Cargo.toml
```diff
@@ -0,0 +1,42 @@
...
+grug-app              = { workspace = true }
+grug-types            = { workspace = true }
+grug-math             = { workspace = true }
+serde_json            = { workspace = true }
```
Rhaki, line 28: sort dependencies alphabetically

indexer/sql/src/active_model.rs
```diff
@@ -0,0 +1,95 @@
...
+#[derive(Debug, Default)]
+pub struct Models {
+    pub block: entity::blocks::ActiveModel,
+    pub transactions: Vec<entity::transactions::ActiveModel>,
+    pub messages: Vec<entity::messages::ActiveModel>,
+    pub events: Vec<entity::events::ActiveModel>,
+}
+
+impl Models {
+    pub fn push(&mut self, tx: &Tx, tx_outcome: &TxOutcome) {
```
Rhaki, line 17: This function can panic. It's called as task with tokio, so only the task should be terminated. Anyway it would be better to return a Result.

Also because this function is called only inside `BlockToIndex::save` function. which already return ` Result<(), sea_orm::DbErr>`

indexer/sql/src/active_model.rs
```diff
@@ -0,0 +1,95 @@
...
+            block_height: self.block.block_height.clone(),
+            hash: Set("".to_string()),
+            data: Set(tx.data.clone().into_inner()),
+            sender: Set(sender),
+            credential: Set(tx.credential.clone().into_inner()),
+        };
+        self.transactions.push(new_transaction);
+
+        for message in tx.msgs.iter() {
+            let serialized_message = serde_json::to_value(message).unwrap();
```
Rhaki, line 36: you can use trait `JsonSerExt` from `grug_types` instead tt maintain consistency with others crates
```suggestion
            let serialized_message = message.to_json_value()?;
```

indexer/sql/src/blocking_indexer.rs
```diff
@@ -0,0 +1,359 @@
...
+    db_txn: Arc<Mutex<Option<(u64, DatabaseTransaction)>>>,
+}
+
+impl BlockingIndexer {
+    pub fn new() -> Result<Self, anyhow::Error> {
+        let runtime = Arc::new(Builder::new_multi_thread()
+                //.worker_threads(4)  // Adjust as needed
+                .enable_all()
+                .build()
+                .unwrap());
```
Rhaki, line 28: use `?` instead `unwrap`

indexer/sql/src/blocking_indexer.rs
```diff
@@ -0,0 +1,359 @@
...
+
+        let db = runtime.block_on(async { Context::connect_db().await })?;
+
+        Ok(BlockingIndexer {
+            context: Context { db },
+            runtime: runtime.clone(),
```
Rhaki, line 34: `clone` is not needed

indexer/sql/src/blocking_indexer.rs
```diff
@@ -0,0 +1,359 @@
...
+            db_txn: Arc::new(Mutex::new(None)),
+        })
+    }
+
+    pub fn new_with_database(database_url: &str) -> Result<Self, anyhow::Error> {
+        let runtime = Arc::new(Builder::new_multi_thread()
+                //.worker_threads(4)  // Adjust as needed
+                .enable_all()
+                .build()
+                .unwrap());
```
Rhaki, line 44: use ? instead unwrap

indexer/sql/src/blocking_indexer.rs
```diff
@@ -0,0 +1,359 @@
...
+
+        let db = runtime.block_on(async { Context::connect_db_with_url(database_url).await })?;
+
+        Ok(BlockingIndexer {
+            context: Context { db },
+            runtime: runtime.clone(),
```
Rhaki, line 50: `clone` is not needed

grug/testing/tests/indexing.rs
penso: I copied that code over from existing test (but looking at it it looks like this was fixed in other code too).

indexer/sql/src/active_model.rs
penso: Yeah good point.

grug/app/src/indexer.rs
```diff
@@ -0,0 +1,57 @@
...
+    pub fn new() -> NullIndexer {
+        NullIndexer {}
+    }
+}
+
+impl Default for NullIndexer {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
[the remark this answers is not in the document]
penso: What's different with this from [dango/app/src/proposal_preparer.rs](https://github.com/left-curve/left-curve/blob/main/dango/app/src/proposal_preparer.rs#L69) ?

indexer/sql/src/non_blocking_indexer.rs
```diff
@@ -0,0 +1,343 @@
...
+}
+
+impl Indexer for NonBlockingIndexer {
+    type Error = crate::error::IndexerError;
+
+    fn start(&mut self) -> error::Result<()> {
+        let context = self.context.clone();
+        block_call(self.runtime.as_ref(), &self.handle, async move {
+            context.migrate_db().await
+        })?;
```
Rhaki, line 190: There is no need to clone `self.context`. `Context.migrate_db()` take `&self` as param, so you can remove `move` keyworks and use `self.context.migrate_db()`.
```suggestion
        block_call(self.runtime.as_ref(), &self.handle, async {
            self.context.migrate_db().await
        })?;
```

indexer/sql/src/non_blocking_indexer.rs
```diff
@@ -0,0 +1,343 @@
...
+        for _ in 0..10 {
+            let blocks = self.blocks.lock().expect("Can't lock blocks");
+            if blocks.is_empty() {
+                break;
+            }
+
+            drop(blocks);
+
+            sleep(Duration::from_millis(10));
+        }
```
Rhaki, line 214: Is possible to to optimize and don't use drop.


```suggestion
        for _ in 0..10 {
            if self.blocks.lock().expect("Can't lock blocks").is_empty() {
                break;
            }
            
            sleep(Duration::from_millis(10));
        }
```

indexer/sql/src/non_blocking_indexer.rs
```diff
@@ -0,0 +1,343 @@
...
+            }
+        }
+
+        Ok(())
+    }
+
+    fn pre_indexing(&self, _block_height: u64) -> error::Result<()> {
+        assert!(self.indexing, "Can't index after shutdown");
+        Ok(())
+    }
```
Rhaki, line 230: If this function return a Result, is better to don't panic here. Returning a result is better cause it could be handled

indexer/sql/src/non_blocking_indexer.rs
```diff
@@ -0,0 +1,343 @@
...
+        assert!(self.indexing, "Can't index after shutdown");
+
+        #[cfg(feature = "tracing")]
+        tracing::info!(block_height = block.height, "index_block called");
+        self.find_or_create(block, |_block_to_index| {
+            #[cfg(feature = "tracing")]
+            tracing::info!(block_height = block.height, "index_block started/finished");
+            Ok(())
+        })
+    }
```
Rhaki, line 242: If this function return a Result, is better to don't panic here. Returning a result is better cause it could be handled

indexer/sql/src/non_blocking_indexer.rs
```diff
@@ -0,0 +1,343 @@
...
+            #[cfg(feature = "tracing")]
+            tracing::info!(block_height = block.height, "index_transaction started");
+
+            block_to_index.txs.push((tx.clone(), tx_outcome.clone()));
+
+            #[cfg(feature = "tracing")]
+            tracing::info!(block_height = block.height, "index_transaction finished");
+            Ok(())
+        })
+    }
```
Rhaki, line 265: The function signature of `index_transaction` require both Tx and TxOutcome to be borrowed. This is not efficient for 2 reason:
1: at line 259 clone both tx and tx_outcome are cloned. At runtime it's needed to read the pointer value and clone it. Is better to move the onwership of both Tx and TxOutcome to this function removing the borrow.
2: This function is called from `App::do_finalize_block` at line `301`. The iterator is created with `into_iter` that properly take the onwership. So tx (and also tx_outcome) can be passed without the borrow without the need to clone them.

indexer/sql/src/non_blocking_indexer.rs
```diff
@@ -0,0 +1,343 @@
...
+
+    fn index_transaction(
+        &self,
+        block: &BlockInfo,
+        tx: &Tx,
+        tx_outcome: &TxOutcome,
+    ) -> error::Result<()> {
+        assert!(self.indexing, "Can't index after shutdown");
```
Rhaki, line 250: If this function return a Result, is better to don't panic here. Returning a result is better cause it could be handled

indexer/sql/src/non_blocking_indexer.rs
```diff
@@ -0,0 +1,343 @@
...
+
+            #[cfg(feature = "tracing")]
+            tracing::info!(block_height = block.height, "index_transaction finished");
+            Ok(())
+        })
+    }
+
+    fn post_indexing(&self, block_height: u64) -> error::Result<()> {
+        assert!(self.indexing, "Can't index after shutdown");
+
```
Rhaki, line 269: If this function return a Result, is better to don't panic here. Returning a result is better cause it could be handled

indexer/sql/src/non_blocking_indexer.rs
```diff
@@ -0,0 +1,343 @@
...
+    {
+        let mut blocks = self.blocks.lock().expect("Can't lock blocks");
+        let block_to_index = blocks.entry(block.height).or_insert(BlockToIndex {
+            block_info: *block,
+            txs: vec![],
+        });
+        action(block_to_index)
+    }
+
+    fn find_or_fail(&self, block_height: &u64) -> error::Result<BlockToIndex> {
```
Rhaki, line 150: `u64` implement `Copy`, don't borrow it
```suggestion
    fn find_or_fail(&self, block_height: u64) -> error::Result<BlockToIndex> {
```

penso: How do you return a `&BlockToIndex` which is a ref to a `Arc<Mutex<HashMap>>`?

indexer/sql/src/non_blocking_indexer.rs
penso: But line 304 does add `tx_outcome` to `tx_outcomes` so it'll be cloned there. But changing it to your like.

indexer/sql/src/non_blocking_indexer.rs
```diff
@@ -0,0 +1,343 @@
...
+impl BlockToIndex {
+    /// Takes care of inserting the data in the database
+    pub async fn save(self, db: &DatabaseTransaction) -> error::Result<()> {
+        #[cfg(feature = "tracing")]
+        tracing::info!(block_height = self.block_info.height, "Indexing block");
+
+        let mut models = Models::build(&self.block_info)?;
+        for tx in self.txs.iter() {
+            models.push(&tx.0, &tx.1)?;
+        }
```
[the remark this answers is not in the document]
penso: yeah good call.

grug/app/src/app.rs
```diff
@@ -1,12 +1,15 @@
+#[cfg(feature = "tracing")]
+use tracing::error;
```
larry0x, line 2: Instead of importing this, let's always invoke tracing macros as

```rust
tracing::error!(/* ... */);
```

Because just `error!` isn't very clear what it means. It can be misunderstood as the function returning an error. With `tracing::error!` it's clear.

grug/app/src/app.rs
```diff
@@ -48,23 +51,25 @@ pub struct App<DB, VM, PP = NaiveProposalPreparer> {
...
     }
 }
 
-impl<DB, VM, PP> App<DB, VM, PP>
+impl<DB, VM, PP, ID> App<DB, VM, PP, ID>
 where
     DB: Db,
     VM: Vm + Clone,
+    ID: Indexer,
     PP: ProposalPreparer,
```
larry0x, line 71: Let's put `ID` under `PP`. Let's keep the order of trait bound the same as the order of generics.

grug/app/src/app.rs
```diff
@@ -505,12 +525,13 @@ where
 // Borsh encoding. This is because these are the methods that clients interact
 // with, and it's difficult to do Borsh encoding in JS client (JS sucks).
 #[cfg(feature = "abci")]
-impl<DB, VM, PP> App<DB, VM, PP>
+impl<DB, VM, PP, ID> App<DB, VM, PP, ID>
 where
     DB: Db,
     VM: Vm + Clone,
+    ID: Indexer,
     PP: ProposalPreparer,
```
larry0x, line 533: Here, `ID` to be below `PP`.

grug/app/src/error.rs
```diff
@@ -8,12 +9,18 @@ pub enum AppError {
...
+    Infallible(#[from] Infallible),
+
     #[error("VM error: {0}")]
     Vm(String),
 
     #[error("DB error: {0}")]
     Db(String),
 
+    #[error("ID error: {0}")]
+    Id(String),
```
larry0x, line 22: Can you rename this to `Indexer`

grug/app/src/indexer.rs
```diff
@@ -0,0 +1,52 @@
...
+
+/// This is a null indexer that does nothing.
+#[derive(Debug, Clone, Default)]
+pub struct NullIndexer;
+
+impl NullIndexer {
+    pub fn new() -> NullIndexer {
+        NullIndexer {}
+    }
+}
```
larry0x, line 15: Do we need this method?

grug/app/src/traits.rs
```diff
@@ -203,3 +205,39 @@ pub trait ProposalPreparer {
...
+/// This is the trait that the indexer must implement. It is used by the Grug core to index blocks
+pub trait Indexer {
+    type Error: ToString;
+
+    /// Called when initializing the indexer, allowing for DB migration if needed
+    fn start(&mut self) -> Result<(), Self::Error> {
+        Ok(())
+    }
+    /// Called when terminating the indexer, allowing for DB transactions to be committed
+    fn shutdown(&mut self) -> Result<(), Self::Error>;
```
larry0x, line 220: Add an empty line between these two methods

grug/testing/Cargo.toml
```diff
@@ -21,9 +22,11 @@ grug-types         = { workspace = true }
 grug-vm-rust       = { workspace = true }
 k256               = { workspace = true }
 rand               = { workspace = true }
+sea-orm            = { workspace = true }
 serde              = { workspace = true }
 tracing            = { workspace = true }
 tracing-subscriber = { workspace = true }
+indexer-sql        = { workspace = true, features = ["tracing"] }
```
larry0x, line 29: Can you order the dependencies alphabetically (move `indexer-sql` up)

grug/testing/src/builder.rs
```diff
@@ -487,16 +513,18 @@ impl<VM, PP, M1, M2, M3> TestBuilder<VM, PP, M1, M2, M3, Undefined<Addr>, Define
...
+    TestBuilder<VM, PP, ID, M1, M2, M3, Defined<Addr>, Defined<TestAccounts>>
 where
     M1: Serialize,
     M2: Serialize,
     M3: Serialize,
     VM: TestVm + Clone,
     PP: ProposalPreparer,
-    AppError: From<VM::Error> + From<PP::Error>,
+    AppError: From<VM::Error> + From<PP::Error> + From<ID::Error>,
+    ID: Indexer,
```
larry0x, line 525: Move `ID` to below `PP` but above `AppError`

indexer/sql-migration/src/lib.rs
```diff
@@ -0,0 +1,13 @@
+pub use sea_orm_migration::prelude::*;
+mod idens;
+
+mod m20220101_000001_create_table;
```
larry0x, line 4: Let's put all the `mod` statements on top, and then all the `use` statements.

indexer/sql/Cargo.toml
```diff
@@ -0,0 +1,32 @@
+[package]
+name = "indexer-sql"
+version       = { workspace = true }
+authors       = { workspace = true }
+edition       = { workspace = true }
+rust-version  = { workspace = true }
+documentation = { workspace = true }
+repository    = { workspace = true }
+license       = { workspace = true }
+categories    = { workspace = true }
```
larry0x, line 10: Align the equal marks

indexer/sql/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+
+[lib]
+path = "src/lib.rs"
+
+[features]
+tracing = ["dep:tracing"]
+
+[dependencies]
+anyhow                = { workspace = true }
+indexer-sql-migration = { workspace = true }
```
larry0x, line 20: Order dependencies alphabetically

grug/app/src/error.rs
```diff
@@ -8,12 +9,18 @@ pub enum AppError {
     #[error(transparent)]
     Std(#[from] StdError),
 
+    #[error(transparent)]
+    Infallible(#[from] Infallible),
```
larry0x, line 13: For infallible, can you see how we use `NaiveError` in `NaiveProposalPreparer`

indexer/sql-migration/src/lib.rs
```diff
@@ -1,8 +1,8 @@
-pub use sea_orm_migration::prelude::*;
 mod idens;
-
 mod m20220101_000001_create_table;
 
+pub use sea_orm_migration::prelude::*;
+
 pub struct Migrator;
```
penso, line 6: This is generated automatically when creating migration, this will be rewritten all the time.

indexer/sql-migration/README.md
```diff
@@ -1,41 +1,59 @@
...
 
 - Generate a new migration file
-    ```sh
-    cargo run -- generate MIGRATION_NAME
-    ```
+
+  ```sh
+  cargo run -- generate MIGRATION_NAME
+  ```
+
```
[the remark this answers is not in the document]
larry0x: Ok... we use [markdownlint](https://github.com/DavidAnson/markdownlint). This file triggers some linter warnings.

PR #317, opened 2024-11-08
feat(grug): Dec DECIMAL_PLACES as generic

Rhaki:


grug/math/src/conversions.rs
```diff
@@ -1,22 +1,29 @@
...
+    crate::{Dec, FixedPoint, Int, MathError, MathResult, Number, NumberConst},
+    std::cmp::Ordering,
+};
 
 // -------------------------------- int -> dec ---------------------------------
 
-impl<U> Int<U>
+pub trait IntoDec<U, const S: u32> {
+    fn checked_into_dec(self) -> MathResult<Dec<U, S>>;
+}
```
larry0x, line 10: Does this have to be a trait? Imo we either use trait for conversions everywhere, or nowhere... for consistency.

grug/math/src/conversions.rs
```diff
@@ -1,22 +1,29 @@
...
+
+impl<U, const S: u32> IntoDec<U, S> for Int<U>
 where
     Self: Number + Copy + ToString,
```
larry0x, line 14: `Copy` is unnecessary, because `Number` already requires `Copy`

PR #318, opened 2024-11-08
feat(dango): `Denom` metadata

### LINKED ISSUE #275: dango: denom metadata
Proposing this schema:

```json
{
    "name": "Dango",
    "symbol": "DNG",
    "description": "native token of dango",
    "decimals": 6
}
```

This is to be stored under a map in bank contract:

```rust
const DENOM_METADATAS: Map<&Denom, DenomMetadata>;
```

A denom's metadata can be set by the namespace owner, or the chain owner.

Rhaki:
closed #275


dango/testing/tests/token_factory.rs
```diff
@@ -239,3 +242,79 @@ fn token_factory() {
...
+            Coins::default(),
+        )
+        .should_succeed();
+
+    // Query metadata
+    let res = suite
+        .query_wasm_smart(contracts.bank, QueryMetadataRequest { denom })
+        .should_succeed();
+
+    assert_eq!(res, metadata);
```
larry0x, line 319: You can use `.should_succeed_and_equal` for this

dango/types/src/bank.rs
```diff
@@ -64,4 +66,14 @@ pub enum QueryMsg {
         start_after: Option<Part>,
         limit: Option<u32>,
     },
+    #[returns(Metadata)]
+    Metadata { denom: Denom },
```
larry0x, line 70: Should also have an enumerative query method:

```rust
Metadatas { start_after: Option<Denom>, limit: Option<u32> }
```

dango/types/src/bank.rs
```diff
@@ -64,4 +66,14 @@ pub enum QueryMsg {
...
+}
+
+#[grug::derive(Serde, Borsh)]
+pub struct Metadata {
+    pub name: NonEmpty<String>,
+    pub symbol: NonEmpty<String>,
+    pub description: String,
```
larry0x, line 77: Let's design the metadata such that it's compatible with Cosmos SDK: https://github.com/cosmos/cosmos-sdk/blob/main/x/bank/proto/cosmos/bank/v1beta1/bank.proto#L99-L118

and chain-registry: https://github.com/cosmos/chain-registry/blob/master/osmosis/assetlist.json

larry0x: Nevermind, their schema actually isn't good... for example there's no `decimals`; instead they have a `denom_units` which is less straightforward.

PR #336, opened 2024-11-19
feat(dango): `SignMode`

Rhaki:
Sport and Margin account can be set to require multiple Signature for authenticate a tx. In particular:

- `Credential` has been renamed to `Signature`;
- `Tx.credential` is now deserialized into `BtreeMap<Hash160, Signature>` (called `Credential` as type now);
- Added a new enum called `SignMode`:
```rust
#[grug::derive(Serde, Borsh)]
pub enum SignMode {
    Single,
    Restricted {
        threshold: u8,
        allowed_keys: BTreeSet<Hash160>,
    },
}
```
- `allowed_keys` is the best solution cause a user could want to don't include some keys that he consider more weak;
- `single::Params` has a new field sign_mode: SignMode.
- Added `ExecuteMsg` variants for `account-factory` that allows to register a new key and update `SignMode` for a sport/margin account.
- TestAccount now support multisign (for test purpose is ok using default one).


larry0x:
replaced by #358

PR #339, opened 2024-11-21
feat(dango): vesting contract

Rhaki:


dango/types/src/vesting.rs
```diff
@@ -0,0 +1,139 @@
...
+        Ok(Schedule {
+            start_time,
+            cliff: self.cliff,
+            vesting: self.vesting,
+        })
+    }
+}
+
+#[grug::derive(Serde, Borsh)]
+pub struct Position<C = Undefined> {
```
larry0x, line 72: This is not the idiomatic use of the builder pattern...

It should be:

```rust
struct Position<C = Undefined<Uint128>>
where
    C: MaybeDefined<Uint128>,
{
    // ...
    pub claimable_amount: C,
}
```

Then, `ClaimablePosition` is to be defined as:

```rust
// not this!
pub type ClaimablePosition = Position<Uint128>;

// this:
pub type ClaimablePosition = Position<Defined<Uint128>>;
```

dango/vesting/src/state.rs
```diff
@@ -0,0 +1,16 @@
+use {
+    dango_types::vesting::Position,
+    grug::{Addr, Counter, IndexedMap, MultiIndex},
+};
+
+pub const POSITION_INDEX: Counter<u64> = Counter::new("position_index", 0, 1);
+
+pub const POSITIONS: IndexedMap<u64, Position, PositionIndexes> =
+    IndexedMap::new("positions", PositionIndexes {
+        user: MultiIndex::new(|_, position| position.user, "positions", "positions_user"),
```
larry0x, line 10: Let's always use singular for a map's namespace (`position` instead of `positions`) and two underscores for the index (`position__user` instead of `position_user`)

dango/vesting/src/state.rs
```diff
@@ -0,0 +1,16 @@
+use {
+    dango_types::vesting::Position,
+    grug::{Addr, Counter, IndexedMap, MultiIndex},
+};
+
+pub const POSITION_INDEX: Counter<u64> = Counter::new("position_index", 0, 1);
```
larry0x, line 6: As a convention, we always name indexes `NEXT_` something, e.g.

- `NEXT_ACCOUNT_INDEX` in dango-account-factory
- `NEXT_PROPOSAL_ID` in dango-account-safe
- `NEXT_SEQUENCE` in dango-auth
- `NEXT_POOL_ID` in dango-amm

larry0x: Also is it necessary to use `u64`?

`u32` should be sufficient.

dango/vesting/src/execute.rs
```diff
@@ -0,0 +1,63 @@
...
+    }
+}
+
+fn create_position(
+    ctx: MutableCtx,
+    user: Addr,
+    schedule: Schedule<Option<Duration>>,
+) -> anyhow::Result<Response> {
+    let amount = ctx.funds.into_one_coin()?;
+    let index = POSITION_INDEX.increment(ctx.storage)?.1;
```
larry0x, line 27: I don't like this `.1` syntax. Let's always do:

```rust
let (_, index) = NEXT_POSITION_INDEX.increment(ctx.storage)?;
```

dango/types/src/vesting.rs
```diff
@@ -0,0 +1,139 @@
...
+    },
+    Claim {
+        idx: u64,
+    },
+}
+
+#[grug::derive(Serde, QueryRequest)]
+pub enum QueryMsg {
+    #[returns(ClaimablePosition)]
+    Position { idx: u64 },
```
larry0x, line 28: I would create a `PositionIndex` alias, similar to `AccountIndex` in dango-account-factory.

dango/vesting/src/query.rs
```diff
@@ -0,0 +1,61 @@
...
+        QueryMsg::Position { idx } => position(ctx, idx)?.to_json_value(),
+        QueryMsg::Positions { start_after, limit } => {
+            positions(ctx, start_after, limit)?.to_json_value()
+        },
+        QueryMsg::PositionsByUser {
+            user,
+            start_after,
+            limit,
+        } => positions_by_user(ctx, user, start_after, limit)?.to_json_value(),
+    }
```
larry0x, line 22: Can you always use this syntax:

```rust
match msg {
    QueryMsg::Something {} => {
        let res = something(ctx.storage)?;
        res.to_json_value()
    },
}
```

I like the formatting better this way...

dango/vesting/src/query.rs
```diff
@@ -0,0 +1,61 @@
...
+
+fn positions(
+    ctx: ImmutableCtx,
+    start_after: Option<u64>,
+    limit: Option<u32>,
+) -> StdResult<BTreeMap<u64, ClaimablePosition>> {
+    let start = start_after.map(Bound::Exclusive);
+    let limit = limit.unwrap_or(DEFAULT_PAGE_LIMIT) as usize;
+    POSITIONS
+        .range(ctx.storage, start, None, grug::Order::Ascending)
```
larry0x, line 39: Instead of `grug::Order::Ascending`, import `use::grug` and just `Order::Ascending` here.

dango/types/src/vesting.rs
```diff
@@ -0,0 +1,139 @@
...
+        if let Some(cliff) = self.schedule.cliff {
+            time = time + cliff;
+
+            ensure!(
+                now >= time,
+                "nothing to claim during cliff phase: end: {:?}, now: {:?}",
+                time,
+                now
+            );
+        }
```
larry0x, line 123: I don't see why throwing error here. If claimable amount is zero then just return zero.

dango/types/src/vesting.rs
```diff
@@ -0,0 +1,139 @@
...
+
+impl Schedule<Option<Timestamp>> {
+    pub fn set_start_time(self, now: Timestamp) -> anyhow::Result<Schedule> {
+        let start_time = if let Some(start_time) = self.start_time {
+            ensure!(
+                start_time >= now,
+                "invalid start time: {:?}, now: {:?}",
+                start_time,
+                now
+            );
```
larry0x, line 57: This incorrect... Vesting start time isn't necessarily bigger than `now`. The starting time is the time that a contributor started working on the project (according to our contractor agreement) which is certainly earlier than `now` (which would be the chain genesis time).

dango/types/src/vesting.rs
```diff
@@ -0,0 +1,139 @@
...
+        start_after: Option<u64>,
+        limit: Option<u32>,
+    },
+}
+
+#[grug::derive(Serde, Borsh)]
+pub struct Schedule<T = Timestamp> {
+    pub start_time: T,
+    pub cliff: Option<Duration>,
+    pub vesting: Option<Duration>,
```
larry0x, line 46: I don't see why these two should be optional. The smart contract should follow our legal contract. In our legal contract there is always a cliff and a vesting period. Let's not over complicate things.

Rhaki: Ok. The scope of this contract was intended to be used for any vesting purpose, not only for contract agreements (the contract can be used by anyone). This is the reason why both cliff and vesting are options. Anyway having just duration without option allow to set a zero duration to skip a phase

dango/types/src/vesting.rs
Rhaki: Yeah ok, this was in order to avoid typos on star_time. But if we need a start_time in the past the ensure can be removed

dango/types/src/vesting.rs
Rhaki: It was just to have a dedicated error instead to have a zero amount error

larry0x:
@Rhaki two more changes needed:

1. a position should have two schedules: vesting and unlocking. at any time, the claimable amount is **the vested amount or unlocked amount, whichever is smaller,** minus the amount already claimed before.

2. add a method to terminate a position. this can only be called by the position's creator. **the amount of tokens the have not been vested is returned to the creator.** the amount that is vested but not yet withdrawn still remains in the position and can be withdrawn (as long as it's unlocked).

these are required by our contractor agreement. you can see my code in mars-vesting on the logic.

dango/vesting/src/execute.rs
```diff
@@ -0,0 +1,60 @@
...
+    ensure!(
+        position.user == ctx.sender,
+        "you don't have the right, O you don't have the right"
+    );
+
+    let claimable_amount = position.compute_claimable_amount(ctx.block.timestamp)?;
+
+    ensure!(!claimable_amount.is_zero(), "nothing to claim");
+
+    position.claimed_amount += claimable_amount;
```
larry0x, line 48: you can use `.checked_add_assign()` here

dango/vesting/src/query.rs
```diff
@@ -0,0 +1,70 @@
...
+
+    POSITIONS
+        .idx
+        .user
+        .prefix(user)
+        .range(ctx.storage, start, None, Order::Ascending)
+        .take(limit)
+        .map(|res| res.map(|(k, v)| (k, v.with_claimable_amount(ctx.block.timestamp))))
+        .collect()
+}
```
larry0x: these functions should be named `query_position`, `query_positions`, `query_positions_by_user` to be consistent with other contracts in the codebase

dango/genesis/src/lib.rs
```diff
@@ -296,6 +301,21 @@ where
...
 
+    let vesting = instantiate(
+        &mut msgs,
+        vesting_code_hash,
+        &vesting::InstantiateMsg {
+            owner: addresses.get(owner).cloned().unwrap(),
+            unlocking_schedule: vesting::Schedule {
+                start_time: None,
+                cliff: Duration::from_seconds(60 * 60 * 24 * 30 * 9),
+                vesting: Duration::from_seconds(60 * 60 * 24 * 30 * 27),
```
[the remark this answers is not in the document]
Rhaki: i've used `Duration::from_weeks(4 * 9)` and `Duration::from_weeks(4 * 27)`

dango/types/src/vesting.rs
```diff
@@ -0,0 +1,183 @@
...
+#[grug::derive(Serde, Borsh)]
+pub struct Config {
+    pub unlocking_schedule: Schedule,
+}
+
+#[grug::derive(Serde, Borsh)]
+pub struct Schedule {
+    pub start_time: Timestamp,
+    pub cliff: Duration,
+    pub vesting: Duration,
```
larry0x, line 67: This should be named `period` instead of vesting, because the unlocking schedule also has a period...

dango/vesting/src/execute.rs
```diff
@@ -0,0 +1,107 @@
...
+        position.user == ctx.sender,
+        "you don't have the right, O you don't have the right"
+    );
+
+    let unlocking_schedule = UNLOCKING_SCHEDULE.load(ctx.storage)?;
+
+    let claimable_amount =
+        position.compute_claimable_amount(ctx.block.timestamp, &unlocking_schedule)?;
+
+    ensure!(!claimable_amount.is_zero(), "nothing to claim");
```
larry0x: We have the `.is_non_zero()` method which I think is more readable.

dango/types/src/vesting.rs
```diff
@@ -0,0 +1,183 @@
...
+    }
+}
+
+#[grug::derive(Serde, Borsh)]
+pub enum VestingStatus {
+    // Position is active. When active, the token stil being distributed
+    Active(Schedule),
+    // Position is terminated.
+    // When terminated, the snapshot of the vested so far is taken and stored
+    Terminated(Uint128),
```
larry0x: Docs should start with a triple slash (`///`) and with proper punctuation at the end (with period `.`).

dango/testing/tests/vesting.rs
```diff
@@ -0,0 +1,554 @@
...
+
+        // Check the balance of the user
+        suite
+            .query_balance(&accounts.relayer, TEST_AMOUNT.denom.clone())
+            .should_succeed_and_equal(initial_balance + Uint128::new(37));
+    }
+}
+
+// Duration for unlock 1 token
+fn epoche(total_duration: Duration, vesting_amount: Uint128) -> Duration {
```
larry0x: Typo (should be `epoch`)

dango/testing/tests/vesting.rs
```diff
@@ -0,0 +1,554 @@
...
+                user: accounts.relayer.address(),
+                schedule: Schedule {
+                    start_time: Duration::from_seconds(0),
+                    cliff: Duration::from_seconds(0),
+                    vesting: Duration::from_seconds(0),
+                },
+            },
+            Coins::default(),
+        )
+        .should_fail_with_error("invalid payment: expecting 1 coins, found 0");
```
larry0x: Instead of putting a string here. we should use

```rust
.should_fail_with_error(StdError::invalid_payment(1, 0));
```

This way, if we change the error message, the tests would still pass.

larry0x: This problem is not limited to vesting, so let's fix them all in a separate PR...

Issue #344, opened 2024-11-26
grug: IBC implementation


larry0x:
We're going to implement IBC Eureka (https://github.com/cosmos/ibc-go/issues/6985) for Grug. Here is how I think the implementation should look like.

Of course my understanding of Eureka is still somewhat limited, so these are subject to changes and corrections.

To start, each light client is a contract, and each app layer protocol (e.g. ICS-20) is a contract. As such, `client_id`'s and `port_id`'s are simply contract addresses (of `Addr` type).

## Clients

The following types and `Message` variants will be added:

```rust
// these are the types that will appear in a transaction...

enum Message {
    CreateClient(MsgCreateClient),
    UpdateClient(MsgUpdateClient),
    FreezeClient(MsgFreezeClient),
}

enum MsgCreateClient {
    Create {
        code_hash: Hash256,
        // `Json` means any serializable data.
        // the contract will deserialize it into the type it expects.
        client_state: Json,
        consensus_state: Json,
        counterparty: Option<Counterparty>,
        // other fields...
    },
    SetCounterparty { /* ... */ },
}

struct MsgUpdateClient {
    pub header: Json,
    // other fields...
}

struct MsgFreezeClient {
    pub misbehavior: Json,
    // other fields...
}
```

```rust
// these are the types that the chain will feed the contract through FFI
// when executing the above messages...

enum MsgIbcClientExecute {
    Update(MsgUpdateClient),
    Freeze(MsgFreezeClient),
}

enum IbcClientQuery {
    VerifyMembership { /* ... */ },
    VerifyNonMembership { /* ... */ },
}

enum IbcClientQueryResponse {
    VerifyMembership(bool),
    VerifyNonMembership(bool),
}
```

Three new wasm entry points will be created:

```rust
#[cfg_attr(not(feature = "library"), grug::export)]
fn ibc_client_create(
    ctx: MutableCtx,
    msg: MsgIbcClientCreate,
) -> Result<Response, Error> { /* ... */ }

#[cfg_attr(not(feature = "library"), grug::export)]
fn ibc_client_execute(
    ctx: MutableCtx,
    msg: MsgIbcClientExecute,
) -> Result<Response, Error> { /* ... */ }

#[cfg_attr(not(feature = "library"), grug::export)]
fn ibc_client_query(
    ctx: ImmutableCtx,
    req: QueryIbcClientRequest,
) -> Result<QueryIbcClientResponse, Error> { /* ... */ }
```

[...]

PR #354, opened 2024-12-02
feat: implemented ics07-tendermint light client

srdtrk:
addresses: #344


larry0x:
@srdtrk Unfortunately this is no longer relevant as we've decided to go with Hyperlane for interop.

PR #357, opened 2024-12-03
refractor(dango): Auth 1 - Keys specific to usernames

Rhaki:


dango/types/src/account_factory/msg.rs
```diff
@@ -75,13 +73,13 @@ pub enum QueryMsg {
...
     },
-    /// Query a key by its hash.
+    /// Query a key by its hash associated to a username.
     #[returns(Key)]
-    Key { hash: Hash160 },
-    /// Enumerate all keys and hashes.
-    #[returns(BTreeMap<Hash160, Key>)]
+    Key { hash: Hash160, username: Username },
+    /// Enumerate all keys.
+    #[returns(BTreeMap<(Username, Hash160), Key>)]
```
larry0x, line 80: > `BTreeMap<(Username, Hash160), Key>`

I don't think this return type would work, because you can't serialize it to JSON. You will get "key must be a string" error when you to `.to_json_value`.

We can do this:

```rust
pub struct QueryKeyResponseItem {
    pub username: Username,
    pub key_hash: Hash160,
    pub key: Key,
}
```

Then use `Vec<QueryKeyResponseItem>` as the return type.

dango/account-factory/src/query.rs
```diff
@@ -115,35 +110,36 @@ fn query_deposits(
...
     let limit = limit.unwrap_or(DEFAULT_PAGE_LIMIT) as usize;
 
     KEYS.range(storage, start, None, Order::Ascending)
         .take(limit)
+        .map(|res| {
+            res.map(|((username, key_hash), key)| QueryKeyResponseItem {
+                username,
+                key_hash,
+                key,
+            })
```
larry0x, line 132: I like to do this:

```rust
let ((username, key_hash), key) = res?;

Ok(QueryKeyResponseItem {
    username,
    key_hash,
    key,
})
```

dango/types/src/account_factory/msg.rs
```diff
@@ -75,13 +73,13 @@ pub enum QueryMsg {
...
     #[returns(Key)]
-    Key { hash: Hash160 },
-    /// Enumerate all keys and hashes.
-    #[returns(BTreeMap<Hash160, Key>)]
+    Key { hash: Hash160, username: Username },
+    /// Enumerate all keys.
+    #[returns(Vec<QueryKeyResponseItem>)]
     Keys {
-        start_after: Option<Hash160>,
+        start_after: Option<(Username, Hash160)>,
```
larry0x, line 82: I like to have a type for this `start_after` as well. Maybe something like

```rust
struct QueryKeyPaginator {
    pub username: Username,
    pub hash: Hash156,
}
```

dango/types/src/account_factory/msg.rs
```diff
@@ -102,7 +100,12 @@ pub enum QueryMsg {
     /// Query a single user by username.
     #[returns(User)]
     User { username: Username },
-    /// Given a key hash, look up the usernames associated with it.
-    #[returns(BTreeSet<Username>)]
-    UsersByKey { hash: Hash160 },
+}
+
+#[grug::derive(Serde)]
+
```
larry0x, line 106: Remove this empty line

PR #359, opened 2024-12-03
feat(Indexer): Add indexer block cache + persistence

penso:
This ensures indexer won't miss a block if process is killed.

- Store data to be indexed on disk cache before indexing
- Delete on disk cache after data was stored on SQL db, or compress and persist it if the `Dango` flag was to persist blocks.
- At start time, existing disk cache with previous block height from RocksDB is indexed
- Remove `index_transaction`, all is happening in `index_block`
- Add `grug_types::Block`


grug/testing/src/suite.rs
```diff
@@ -223,13 +223,38 @@ where
...
         genesis_block: BlockInfo,
         genesis_state: GenesisState,
     ) -> Self {
+        // This is doing the same order as in Dango.
+        // 1. Calling `start` on the indexer
+
+        let previous_block_height = match genesis_block.height {
+            0..=1 => None,
+            2.. => Some(genesis_block.height - 1),
+        };
```
Rhaki, line 239:
```suggestion
        let previous_block_height = if let 0 | 1 = genesis_block.height {
            None
        } else {
            Some(genesis_block.height - 1)
        };
```

grug/types/src/app.rs
```diff
@@ -97,6 +97,13 @@ pub struct BlockInfo {
     pub hash: Hash256,
 }
 
+#[derive(Serialize, Deserialize, BorshSerialize, BorshDeserialize, Debug, Clone, PartialEq, Eq)]
+#[serde(deny_unknown_fields)]
+pub struct Block {
+    pub block_info: BlockInfo,
+    pub txs: Vec<Tx>,
+}
```
Rhaki, line 105: If possible try to avoid to create this container struct. We already have a lot of types, create new types only if needed (in this case Block is only used to wrap `block_info` and `txs`, not a big deal)

penso: I only did as @larry0x suggested at https://discord.com/channels/1246466309080879207/1260999513859166342/1316133794734149712:

```
Larry Engineer: we should do
struct Block {
    pub info: BlockInfo,
    pub txs: Vec<Tx>,
}
```

Should I have added it somewhere else?

Rhaki: Ok so

indexer/sql/src/non_blocking_indexer.rs
```diff
@@ -192,15 +198,122 @@ impl NonBlockingIndexer {
...
+        })?;
+        Ok(())
+    }
+}
+
+impl NonBlockingIndexer {
+    /// Where will this block be temporarily saved on disk
+    pub fn block_filename(&self, block_height: u64) -> PathBuf {
+        let directory = self.indexer_path.block_path();
+        let filename = format!("{block_height}");
```
Rhaki, line 234: is possible to use `block_height.to_string()`

indexer/sql/src/non_blocking_indexer.rs
```diff
@@ -192,15 +198,122 @@ impl NonBlockingIndexer {
...
+        };
+
+        for block_height in last_indexed_block_height..=latest_block_height {
+            let block_filename = self.block_filename(block_height);
+
+            let block_to_index = match BlockToIndex::load_from_disk(block_filename.clone()) {
+                Ok(block_to_index) => block_to_index,
+                Err(err) => {
+                    #[cfg(feature = "tracing")]
+                    tracing::error!(error = %err, block_height, "can't load block from disk");
```
Rhaki, line 256: is % a type error? do you mean $?

indexer/sql/src/non_blocking_indexer.rs
```diff
@@ -291,29 +391,44 @@ impl Indexer for NonBlockingIndexer {
...
 
         let context = self.context.clone();
         let block_to_index = self.find_or_fail(block_height)?;
         let blocks = self.blocks.clone();
+        let keep_blocks = self.keep_blocks;
+        let block_filename = self.block_filename(block_to_index.block.block_info.height);
 
         // NOTE: I can't remove the block to index *before* indexing it with DB txn committed, or
         // the shutdown method could be called and see no current block being indexed, and quit.
         // The block would then not be indexed.
```
Rhaki, line 405:
```suggestion
```

indexer/sql/src/non_blocking_indexer.rs
```diff
@@ -192,15 +198,122 @@ impl NonBlockingIndexer {
...
+impl NonBlockingIndexer {
+    /// Where will this block be temporarily saved on disk
+    pub fn block_filename(&self, block_height: u64) -> PathBuf {
+        let directory = self.indexer_path.block_path();
+        let filename = format!("{block_height}");
+        directory.join(filename)
+    }
+
+    /// Load all existing tmp files and ensure they've been indexed
+    fn ensure_previous_indexed_blocks(&self, latest_block_height: u64) -> error::Result<()> {
```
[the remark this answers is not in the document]
penso: Renamed to `index_previous_unindexed_blocks`

> Also, what happend if we remove from the db a block that is the latest indexed? That block is not reindexed again right?

No block should ever be removed from the DB until we implement an indexer prune which nodes could run to clean up space. It's only removed in the test to build the context of that test.

To reindex missing blocks, you need to save blocks on disk with `set_keep_blocks()` and I'll need to write a binary (coming later) to "reindex" all on-disk blocks. This will be needed once we're in production and feature changes in the indexer code requires to reindex previous blocks. It's rare, but it will happen and so why I have that `keep_blocks` feature.

Cargo.toml
```diff
@@ -143,6 +144,7 @@ dango-vesting         = { path = "dango/vesting" }
 # Indexer packages
 indexer-sql           = { path = "indexer/sql" }
 indexer-sql-migration = { path = "indexer/sql-migration" }
+indexer-disk-saver    = { path = "indexer/disk-saver" }
```
larry0x, line 147: Sort these alphabetically

grug/app/src/home.rs
```diff
@@ -0,0 +1,22 @@
...
+    /// Used for the RocksDB database
+    pub fn data_dir(&self) -> PathBuf {
+        self.home.join("data")
+    }
+
+    /// Used for the indexer, used to store blocks before they're saved to the DB
+    pub fn indexer_dir(&self) -> PathBuf {
+        self.home.join("indexer")
+    }
+}
```
larry0x, line 22: Must this be in grug-app? I think filesystem should be the concern of the CLI, not the app.

grug/types/src/app.rs
```diff
@@ -97,6 +97,13 @@ pub struct BlockInfo {
     pub hash: Hash256,
 }
 
+#[derive(Serialize, Deserialize, BorshSerialize, BorshDeserialize, Debug, Clone, PartialEq, Eq)]
+#[serde(deny_unknown_fields)]
+pub struct Block {
+    pub block_info: BlockInfo,
```
larry0x, line 103: Can you rename this field to just `info`

indexer/disk-saver/Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+tempfile              = { workspace = true }
+thiserror             = { workspace = true }
+tokio                 = { workspace = true }
+tracing               = { workspace = true, optional = true }
+uuid                  = { workspace = true }
+
+[dev-dependencies]
+assertor              = { workspace = true }
+tracing-subscriber    = { workspace = true }
+tracing               = { workspace = true }
```
larry0x, line 39: Sort alphabetically (tracing-subscriber after tracing)

indexer/disk-saver/src/error.rs
```diff
@@ -0,0 +1,19 @@
...
+    #[error(transparent)]
+    Io(#[from] std::io::Error),
+
+    #[error(transparent)]
+    Persist(#[from] tempfile::PersistError),
+
+    #[error(transparent)]
+    Lzma(#[from] lzma_rs::error::Error),
+
+    #[error("File already exists: {0}")]
```
larry0x, line 17: Do you uppercase the error message. See [rust style guide](https://rustc-dev-guide.rust-lang.org/diagnostics.html#diagnostic-output-style-guide:~:text=Error%2C%20Warning%2C%20Note%2C%20and%20Help%20messages%20start%20with%20a%20lowercase%20letter%20and%20do%20not%20end%20with%20punctuation.)

indexer/disk-saver/src/persistence.rs
```diff
@@ -0,0 +1,248 @@
...
+            let mut compressed = Vec::new();
+            lzma_compress(&mut serialized.as_slice(), &mut compressed)?;
+            tmp_file.write_all(&compressed)?;
+        } else {
+            tmp_file.write_all(&serialized)?;
+        }
+
+        tmp_file.flush()?;
+        tmp_file.persist(&self.file_path)?;
+        Ok(())
```
larry0x, line 98: Add an empty line before `Ok(())`. In general I like to have an empty line before the return value (except for the very small functions)

indexer/disk-saver/src/persistence.rs
```diff
@@ -0,0 +1,248 @@
...
+        std::fs::remove_file(&self.file_path)?;
+        Ok(())
+    }
+
+    pub fn delete_file_path(file_path: &PathBuf) -> Result<(), Error> {
+        std::fs::remove_file(file_path)?;
+        Ok(())
+    }
+}
+
```
larry0x, line 185: Always add this line to separate logics from tests:

```rust
// ----------------------------------- tests -----------------------------------
```

indexer/sql/src/block_to_index.rs
```diff
@@ -0,0 +1,149 @@
...
+        let mut block_to_index: Self = DiskPersistence::new(file_path.clone(), false).load()?;
+        block_to_index.filename = file_path;
+        Ok(block_to_index)
+    }
+
+    pub fn delete_from_disk(file_path: PathBuf) -> error::Result<()> {
+        Ok(DiskPersistence::new(file_path, false).delete()?)
+    }
+}
+
```
larry0x, line 107: Add separator comment

PR #367, opened 2024-12-06
fix(grug): Crono not revert state changes on error

Rhaki:
Crono execution was not reverted if an error happens.

### DIFF
--- a/grug/app/src/execute.rs
+++ b/grug/app/src/execute.rs
@@ -2,8 +2,8 @@ use {
     crate::{
         call_in_0_out_1_handle_response, call_in_1_out_1, call_in_1_out_1_handle_response,
         call_in_2_out_1_handle_response, handle_response, has_permission, schedule_cronjob, AppCtx,
-        AppError, AppResult, MeteredItem, MeteredMap, Vm, APP_CONFIG, CODES, CONFIG, CONTRACTS,
-        NEXT_CRONJOBS,
+        AppError, AppResult, Buffer, MeteredItem, MeteredMap, Shared, Vm, APP_CONFIG, CODES,
+        CONFIG, CONTRACTS, NEXT_CRONJOBS,
     },
     grug_math::Inner,
     grug_types::{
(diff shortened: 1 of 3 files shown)


larry0x:
@Rhaki when you have time can you sync this up with the latest main branch? on main branch we're no longer using `AppCtx`

grug/app/src/execute.rs
```diff
@@ -777,16 +777,22 @@ where
...
 
-pub fn do_cron_execute<VM>(ctx: AppCtx<VM>, contract: Addr) -> AppResult<Vec<Event>>
+pub fn do_cron_execute<VM>(mut ctx: AppCtx<VM>, contract: Addr) -> AppResult<Vec<Event>>
 where
     VM: Vm + Clone,
     AppError: From<VM::Error>,
 {
+    let buffer = Shared::new(Buffer::new(ctx.storage, None));
+
+    ctx.storage = Box::new(buffer.clone());
```
larry0x, line 787: stuff like this is why i decided to get rid of `AppCtx`. mutating `ctx.storage` just feel wrong... i.e. an anti-pattern

grug/testing/tests/cron.rs
```diff
@@ -36,6 +41,18 @@ mod tester {
...
+    pub fn failing_instantiate(_: MutableCtx, _: Empty) -> StdResult<Response> {
+        Ok(Response::new())
+    }
+
+    pub fn failing_cron_execute(ctx: SudoCtx) -> StdResult<Response> {
+        SHOULD_BE_EMPTY.save(ctx.storage, &()).unwrap();
+
+        Err(StdError::host(
+            "This cronjob is supposed to fail".to_string(),
+        ))
```
larry0x, line 54: `StdError::host` means the host returned an error, so this is a misuse here. you can just use `anyhow::Result` for this function.

grug/testing/tests/cron.rs
```diff
@@ -36,6 +41,18 @@ mod tester {
 
         Ok(Response::new().add_message(Message::transfer(job.receiver, job.coin).unwrap()))
     }
+
+    pub fn failing_instantiate(_: MutableCtx, _: Empty) -> StdResult<Response> {
+        Ok(Response::new())
+    }
```
larry0x, line 47: instead of this function, i would create a new `mod`:

```rust
mod failing_tester {
    pub fn instantiate() { /* ... */ }

    pub fn cron_execute() { /* ... */ }
}
```

one mod, one function. a lot more logical this way

grug/testing/tests/cron.rs
```diff
@@ -36,6 +41,18 @@ mod tester {
...
+
+    pub fn failing_cron_execute(ctx: SudoCtx) -> StdResult<Response> {
+        SHOULD_BE_EMPTY.save(ctx.storage, &()).unwrap();
```
larry0x, line 50: instead of creating a new `Item`, it's easier to just do `ctx.storage.set(b"foo", b"bar")`

Rhaki:
Closed in favor of #408

PR #376, opened 2024-12-11
refactor(grug): reorganize some contents in app and types

larry0x:
Split big files into smaller files for better readability.


grug/app/src/execute/finalize.rs
```diff
@@ -0,0 +1,94 @@
...
+    #[cfg(feature = "tracing")]
+    evt.debug(
+        |_| {
+            tracing::info!(sender = tx.sender.to_string(), "Finalized fee");
+        },
+        "Failed to finalize fee",
+        // `finalize_fee` is supposed to always succeed, so if it doesn't,
+        // we print a tracing log at ERROR level to highlight the seriousness.
+        // #[cfg(feature = "tracing")]
+        // tracing::error!(err = err.to_string(), "Failed to finalize fee")
```
larry0x, line 33: hmm, here we use `tracing::warn!` but it should really be `tracing::error!`

PR #379, opened 2024-12-12
docs: pool limit order and  XYK pool math

Zexsor33:


larry0x:
Sorry @Zexsor33 this is no longer relevant as we're going for orderbook now

PR #402, opened 2024-12-29
feat(dango): orderbook

larry0x:
TODOs:

- [x] single contract for all pairs (?)
- [ ] listing / pair creation fee
- [ ] trading fee
- [x] batch send
- [x] events
- [ ] ~market orders~ (leave for a later PR)
- [ ] ~DCA~ (leave for a later PR)
- [ ] ~TWAP~ (not necessary)


grug/types/src/coins.rs
```diff
@@ -191,33 +191,33 @@ impl Coins {
     }
 
     /// Insert a new coin to the `Coins`.
-    pub fn insert(&mut self, coin: Coin) -> StdResult<()> {
+    pub fn insert(&mut self, coin: Coin) -> StdResult<&mut Self> {
```
larry0x, line 194: put `&mut Self` in the return so that i can chain multiple `.insert` methods. the syntax looks a little better this way.

grug/app/src/app.rs
```diff
@@ -190,6 +190,14 @@ where
...
     }
 
+    // Finalize a block by performing the following actions in order:
+    //
+    // 1. indexer `pre_indexing`
+    // 2. execute transactions one by one
+    // 3. perform cronjobs
+    // 4. remove orphaned nodes
+    // 5. flush (but not commit) state changes to DB
+    // 5. indexer `index_block`
```
larry0x, line 200: move cronjobs after transactions. this way, if user submits an order, and there's a match, then the order is filled **in the same block** where it's submitted. otherwise, user has to wait until the next block for the order to be filled.

dango/testing/tests/amm.rs
```diff
@@ -22,7 +22,7 @@ static LP_2: LazyLock<Denom> = LazyLock::new(|| Denom::from_str("amm/pool/2").un
 
 #[test]
 fn amm() {
-    let (mut suite, mut accounts, _, contracts) = setup_test();
+    let (mut suite, mut accounts, _, contracts) = setup_test_naive();
```
larry0x, line 25: these tests don't need the Pyth ProposalPreparer, so i changed to naive PP

grug/testing/tests/cron.rs
```diff
@@ -143,26 +143,34 @@ fn cronjob_works() {
...
             uatom: 2,
+            uosmo: 0,
+            umars: 0,
+        },
+        // Block time: 7
+        //
+        // cron1 sends 1 uatom, rescheduled to 8 (it runs out of coins here)
+        // cron2 sends 1 uosmo, rescheduled to 9
+        Balances {
+            uatom: 3,
```
larry0x, line 167: now we move cronjob to end of block, the behavior of this test case changes & needs to be updated

larry0x:
a few points that need to be worked on:

- **rationing:** currently, in case there's excess demand on one side of the book, then the side with more demand can only be filled partially (e.g. 100 base asset of BUY and 120 base asset of SELL, then the 20 SELL won't be filled.) in this case we follow price-time priority. the more proper way is: orders submitted in earlier blocks have priority; orders submitted in the same block are to be rationed.

- **sealed bid:** orders submitted during the block shouldn't be inserted into the book just yet. if we do, contracts would be able to query the orders during the block. instead the new orders should be kept in a "transient storage" that can't be seen by other contracts. when doing matching in `cron_execute`, the transient storage is merged into the main storage.

- **market orders:** following Injective's implementation, market orders are to be handled separately from limit orders, the this flow during `cron_execute`:

  - fill market orders (without merging the new orders into the book yet)
  - merge new orders in the transient storage into the book
  - fill limit orders

dango/orderbook/src/execute.rs
```diff
@@ -0,0 +1,326 @@
...
+    },
+    grug::{
+        Addr, Coin, Coins, ContractEvent, Denom, Message, MultiplyFraction, MutableCtx, Number,
+        Order as IterationOrder, Response, StdResult, Storage, SudoCtx, Udec128, Uint128,
+    },
+    std::collections::{BTreeMap, BTreeSet},
+};
+
+// equals 0.5
+const HALF: Udec128 = Udec128::raw(Uint128::new(500_000_000_000_000_000));
```
Rhaki, line 22:
```suggestion
const HALF: Udec128 = Udec128::new_percent(50);
```

larry0x: can you also delete the "equals 0.5" comment

apollo-sturdy (commented):
One issue I see with this design is atomicity of trading for DeFi contracts. Perhaps it's not needed for now, but many DeFi apps have a trade as part of the execution and then proceeds to do something with the bought assets. Since orders are matched at the end of the block, this would prevent these kinds of apps from working. Any thoughts on this?

apollo-sturdy:
> One issue I see with this design is atomicity of trading for DeFi contracts. Perhaps it's not needed for now, but many DeFi apps have a trade as part of the execution and then proceeds to do something with the bought assets. Since orders are matched at the end of the block, this would prevent these kinds of apps from working. Any thoughts on this?

Perhaps we could add a market order type that causes order matching straight away, rather than in the cronjob?

larry0x:
> One issue I see with this design is atomicity of trading for DeFi contracts.

@apollo-sturdy Two solutions I see:

1. The trader can specify a hook message as part of the trade. Once the trade is filled (partially or completely), the trader's contract is called.

   For this to work we need a gas limit for the submessages, because order clearing happens as an end-block cronjob which isn't subject to gas, so the hooks need to be limited in gas.

2. Some trading pairs (mostly long tail, small cap assets) will have XYK passive liquidity pools (I'm writing a document on this, check discord general channel). Defi protocol can directly swap in this pool synchronously.

larry0x:
Things to do to prevent spamming:

- permissioned creation of trading pairs
- minimum order size

dango/orderbook/src/execute.rs
```diff
@@ -0,0 +1,325 @@
...
+/// Implemented according to:
+/// <https://motokodefi.substack.com/p/uniform-price-call-auctions-a-better>
+#[cfg_attr(not(feature = "library"), grug::export)]
+pub fn cron_execute(ctx: SudoCtx) -> StdResult<Response> {
+    let mut events = Vec::new();
+    let mut refunds = BTreeMap::new();
+
+    // Find all pairs that have received new orders during the block.
+    let pairs = NEW_ORDER_COUNTS
+        .current_range(ctx.storage, None, None, IterationOrder::Ascending)
```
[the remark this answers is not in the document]
larry0x: This runs as an end-block cronjob which has unlimited gas.

To prevent spamming we can do:

- permissioned creation of trading pairs
- minimum order size

dango/orderbook/src/execute.rs
```diff
@@ -0,0 +1,325 @@
...
+            );
+
+            ensure!(
+                deposit.amount == amount,
+                "incorrect deposit amount for SELL order! expecting: {}, found: {}",
+                amount,
+                deposit.amount
+            );
+        },
+    }
```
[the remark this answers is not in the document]
larry0x: For this problem, I think frontend prevention is sufficient.

But for other reasons (spamming prevention) we should have permissioned pair creation, meaning traders can only place orders in a predefined set of {base, quote} pairs.

dango/types/src/orderbook.rs
```diff
@@ -0,0 +1,191 @@
...
+    ///   ```plain
+    ///   ceil(amount * price)
+    ///   ```
+    SubmitOrder {
+        base_denom: Denom,
+        quote_denom: Denom,
+        direction: Direction,
+        amount: Uint128,
+        price: Udec128,
+    },
```
Rhaki, line 116: amount can be derived by the price * coin.amount in case of a BID and coin.amount in case of ASK.
This works like an extra check in case of BID but not mandatory

larry0x: We do the reverse, compute and assert `coin.amount` based on `amount` and `price`

PR #408, opened 2025-01-02
fix(grug): Crono not revert state changes on error

Rhaki:


grug/app/src/app.rs
```diff
@@ -263,25 +263,32 @@ where
...
                 time,
                 next_time,
             );
 
+            // Commit state changes if the cronjob was successful.
+            // Ignore if unsuccessful.
+            if cron_event.is_ok() {
+                cron_buffer.disassemble().commit();
+            }
+
```
larry0x, line 285: @Rhaki I moved the buffer commit logic from `execute/cron.rs` to here in `app.rs`. This is more consistent with the rest of the codebase.

PR #409, opened 2025-01-02
fix(dango): safe account auth issues

### LINKED ISSUE #262: bug: voter authentication in Safe account

larry0x:
closes #262 #263

replaces #267

### DIFF
--- a/dango/account/safe/src/execute.rs
+++ b/dango/account/safe/src/execute.rs
@@ -62,6 +62,11 @@ pub fn receive(_ctx: MutableCtx) -> StdResult<Response> {
 
 #[cfg_attr(not(feature = "library"), grug::export)]
 pub fn execute(ctx: MutableCtx, msg: ExecuteMsg) -> anyhow::Result<Response> {
+    ensure!(
+        ctx.sender == ctx.contract,
+        "only the Safe account itself can execute itself"
+    );
+
     match msg {
         ExecuteMsg::Propose {
             title,
--- a/dango/account/safe/src/execute.rs
+++ b/dango/account/safe/src/execute.rs
@@ -1,7 +1,7 @@
 use {
     crate::{NEXT_PROPOSAL_ID, PROPOSALS, VOTES},
     anyhow::{bail, ensure},
-    dango_auth::authenticate_tx,
+    dango_auth::verify_nonce_and_signature,
     dango_types::{
         account::{
             multi::{ExecuteMsg, Proposal, ProposalId, Status, Vote},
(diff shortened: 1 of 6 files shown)


dango/account/safe/src/execute.rs
```diff
@@ -39,18 +39,41 @@ pub fn authenticate(ctx: AuthCtx, tx: Tx) -> anyhow::Result<AuthResponse> {
...
+                    }
                 }
             },
-            _ => bail!("a Safe account can only execute itself"),
+            _ => bail!("the only action a Safe account can do is to execute itself"),
         }
     }
 
-    authenticate_tx(ctx, tx, None, Some(metadata))?;
+    verify_nonce_and_signature(ctx, tx, None, Some(metadata))?;
```
larry0x, line 76: This addresses #262

dango/account/safe/src/execute.rs
```diff
@@ -62,6 +85,11 @@ pub fn receive(_ctx: MutableCtx) -> StdResult<Response> {
 
 #[cfg_attr(not(feature = "library"), grug::export)]
 pub fn execute(ctx: MutableCtx, msg: ExecuteMsg) -> anyhow::Result<Response> {
+    ensure!(
+        ctx.sender == ctx.contract,
+        "only the Safe account itself can execute itself"
+    );
```
larry0x, line 91: This addresses #263

dango/account/safe/src/execute.rs
```diff
@@ -334,33 +361,6 @@ mod tests {
...
-                }
-                .to_json_value()
-                .unwrap(),
-                credential: Json::null(),
-            });
-
-            assert!(res.is_err_and(|err| err.to_string().contains(&format!(
-                "account {SAFE} isn't associated with user `{non_member}`"
-            ))));
-        }
```
larry0x, line 362: This test case is no longer relevant as the same scenario is already covered by the integration test (in dango-testing)

dango/auth/src/lib.rs
```diff
@@ -66,6 +61,40 @@ pub fn authenticate_tx(
...
+///
+/// This is intended for the multi-signature (Safe) account, where we ensure the
+/// username is associated with the Safe _at the time a proposal was created_,
+/// instead of _now_.
+pub fn verify_nonce_and_signature(
+    ctx: AuthCtx,
+    tx: Tx,
+    maybe_factory: Option<Addr>,
+    maybe_metadata: Option<Metadata>,
+) -> anyhow::Result<()> {
```
larry0x, line 80: My solution to #263 is to split `authenticate_tx` into two functions.

Spot and Margin accounts use `authenticate_tx`. This function:

1. ensures signer username is associated with the sender account
2. ensures nonce is correct
3. ensures signature is valid

Then a new function `verify_nonce_and_signature` is used by Safe account. This function skips (1) and only does (2) and (3).

larry0x (commented):
I also pretty much rewrote all the tests in `dango/testing/tests/safe.rs`

PR #413, opened 2025-01-04
lint(grug): more fixes related to optional dependencies in grug-app

larry0x:
use the following commands to make sure:

```
cargo clippy -p grug-app --no-default-features
cargo clippy -p grug-app --no-default-features --features abci
cargo clippy -p grug-app --no-default-features --features tracing
cargo clippy -p grug-app --no-default-features --features abci,tracing
```

### DIFF
--- a/grug/app/Cargo.toml
+++ b/grug/app/Cargo.toml
@@ -10,7 +10,7 @@ rust-version  = { workspace = true }
 version       = { workspace = true }
 
 [features]
-abci    = ["data-encoding", "tower", "tower-abci"]
+abci    = ["data-encoding", "tendermint", "tower", "tower-abci"]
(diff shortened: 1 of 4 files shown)


grug/app/Cargo.toml
```diff
@@ -10,7 +10,7 @@ rust-version  = { workspace = true }
 version       = { workspace = true }
 
 [features]
-abci    = ["data-encoding", "tower", "tower-abci"]
+abci    = ["data-encoding", "tendermint", "tower", "tower-abci"]
```
larry0x, line 13: technically, data-encoding should only be enabled if both `abci` and `tracing` are enabled. but i don't know to do it. in practice though, if we use `abci` we always use `tracing` so not a problem.

PR #414, opened 2025-01-06
feat(dango): track seen nonces

j0nl1:


dango/types/src/account/spot.rs
```diff
@@ -1,7 +1,9 @@
...
+
 /// Query messages for the spot account
 #[grug::derive(Serde, QueryRequest)]
 pub enum QueryMsg {
-    /// Query the account's current nonce number.
-    #[returns(u32)]
-    Nonce {},
+    /// Query the most recent transaction nonces that have been recorded.
+    #[returns(BTreeSet<Nonce>)]
+    SeenNonces {},
```
larry0x, line 8: @j0nl1 I changed the query function name and the return type here

PR #422, opened 2025-01-20
doc: dango dex paper

larry0x:


pacmanifold (commented):
mostly nitpick spelling corrections. some questions on the math of the passive pool. 

## Thoughts on the implementation of the passive pool

### Thought nr. 1
I was just thinking what happens with the distribution if there is a jump in oracle price and the impact on the inventory of the pool. The image shows the order placement if the oracle price would jump 10%. 
Screenshot 2025-01-24 at 14 15 51

As can be seen from the order depth on the ask side the total SOL needed is ~0.394 as compared to before price gap when total SOL needed was 0.363. I can think of three approaches here
1. Cut the tail of the order distribution so total order depth doesn't exceed 0.363
    - Maybe makes the pool vulnerable to being completely traded through
2. Rebalance the portfolio 
    - not sure if this can be done easily through the DEX since there is no guarantee of liquidity
    - this would incur volatility decay over time
3. Only use parts of the liquidity to place orders and keep a reserve for cases like this

### Though nr. 2
What do we do if the oracle is attacked or not fast enough and the pool ends up being entirely or partially traded through and left off side by the market? Do we simply rebalance?

larry0x:
@pacmanifold on nr.1: if pool reserve ratio is 1:200 but oracle price jumps to 230, then the pool would want to adjust its asset ratio also to 1:230. This means it needs to sell SOL and buy USD. Therefore it's seems reasonable it puts more sell orders and buy orders. Thoughts?

On nr.2: the pool should have a "max oracle age" parameter. If oracle price is too old, it should return empty iterators for the orders, meaning not putting any orders in the book.

larry0x:
Merging the first draft now. We may make revisions later in future PRs.

PR #425, opened 2025-01-21
feat(grug): Balance Tracker

### LINKED ISSUE #405: test: balance tracker
for tests, we often needs to check whether a user's balance has changed with a specific amount before and after a transaction. to simplify this, we can introduce a new primitive `BalanceTracker`, to be included in the `TestSuite` as `suite.balance`.

```rust
// record balances before a block
suite.record_balances(&accounts);

// make a new block
suite.make_block(/* ... */);

// check balance changes after a block
suite.balance.of(&accounts.user1, &denom).should_be_zero();
suite.balance.of(&accounts.user1, &denom).should_be_non_zero();
suite.balance.of(&accounts.user1, &denom).should_be(amount);
suite.balance.of(&accounts.user1, &denom).should_be_around(amount, max_error);
suite.balance.of(&accounts.user1, &denom).should_have_increased_by(amount);
suite.balance.of(&accounts.user1, &denom).should_have_decreased_by(amount);
```

Rhaki:
close #405


dango/testing/tests/dex.rs
```diff
@@ -332,7 +258,9 @@ fn dex_works(
         .collect::<BTreeMap<_, _>>();
 
     // Track the users' balances.
-    tracker.record_balances(&suite, users_by_order_id.values().copied());
+    suite
+        .balance_tracker()
+        .record_balances(users_by_order_id.values().copied());
```
larry0x, line 263: @Rhaki It would be nice if we can directly invoke:

```rust
suite.record_balances(/* ... */);
```

Instead of having to do `.balance_tracker()` first

Rhaki: I originally thought so too. The problem is that trying to use the direct methods in `TestSuite ` created confusion. For a developer who is new to the `TestSuite `, it is not clear that this method modifies (such as cleaning, resetting, adding) the balance tracker and not the onchain balances. Instead, having the dedicated section accessible via the ` balance_tracker()` function, this type of confusion is avoided

PR #429, opened 2025-01-23
test(dango): add genesis config and constants

larry0x:
Currently our test code is somewhat low quality because:

1. The `grug_genesis::build_genesis` function takes some parameters as function inputs while having others hard coded in the function body.

   This PR creates a `GenesisConfig` struct as the function's input so everything is in the input.

3. Tests use a lot of "magic values" such as string literals (e.g. `"uusdc"`). And these are not consistent across tests: some use `"bridge/btc"` while some use `"hyp/bitcoin/sat"` for example.

   This PR creates a `grug_types::constants` module as a single source of truth for all const values.

These changes should allow us to write tests more efficiently.


dango/testing/tests/lending.rs
```diff
@@ -41,21 +35,7 @@ fn feed_oracle_usdc_price(
     accounts: &mut TestAccounts,
     contracts: &Contracts,
 ) {
-    let id = PythId::from_str(USDC_USD_ID).unwrap();
     let precision = 6;
```
larry0x, line 38: Is this accurate? Pyth's website says USDC has an "exponent" of -8. Is precision the token's decimals or the Pyth exponent?

dango/testing/tests/lending.rs
```diff
@@ -41,21 +35,7 @@ fn feed_oracle_usdc_price(
...
-    suite
-        .execute(
-            &mut accounts.owner,
-            contracts.oracle,
-            &dango_types::oracle::ExecuteMsg::RegisterPriceSources(btree_map! {
-                usdc_denom.clone() => dango_types::oracle::PriceSource::Pyth { id, precision }
-            }),
-            Coins::default(),
-        )
-        .should_succeed();
```
larry0x, line 58: We now register all price sources during genesis n `setup_test()`, so no need to do it here.

dango/testing/tests/margin.rs
```diff
@@ -89,27 +84,6 @@ fn assert_approx_eq(a: Udec128, b: Udec128, max_rel_diff: &str) -> Result<(), Te
...
-        .execute(
-            &mut accounts.owner,
-            contracts.oracle,
-            &dango_types::oracle::ExecuteMsg::RegisterPriceSources(btree_map! {
-                denom => dango_types::oracle::PriceSource::Pyth { id, precision }
-            }),
-            Coins::default(),
-        )
-        .should_succeed();
-}
```
larry0x, line 111: Same here, no need to register price sources here

dango/testing/tests/margin.rs
```diff
@@ -284,45 +253,33 @@ fn setup_margin_test_env(
...
-    suite
-        .execute(
-            &mut accounts.owner,
-            contracts.lending,
-            &lending::ExecuteMsg::UpdateMarkets(btree_map! {
-                BTC.clone() => MarketUpdates {},
-            }),
-            Coins::new(),
-        )
-        .should_succeed();
```
larry0x, line 309: Lending markets are now also set up during genesis in `setup_tests()`

dango/testing/tests/margin.rs
```diff
@@ -284,45 +253,33 @@ fn setup_margin_test_env(
...
         .should_succeed();
 
     // Whitelist WBTC as collateral at 80% power
     set_collateral_power(
         suite,
         accounts,
-        BTC.clone(),
+        WBTC_DENOM.clone(),
         CollateralPower::new(Udec128::new_percent(80)).unwrap(),
     );
```
larry0x, line 284: In a future PR, this can also be set up in genesis

dango/types/src/constants/tokens.rs
```diff
@@ -0,0 +1,34 @@
+use {
+    grug::Denom,
+    std::{str::FromStr, sync::LazyLock},
+};
+
+pub static ATOM_DENOM: LazyLock<Denom> =
+    LazyLock::new(|| Denom::from_str("hyp/atom/atom").unwrap());
```
larry0x, line 7: I really hate how rustfmt formats this

dango/testing/tests/lending.rs
Rhaki: pyth expo is different from token precision.
pyth expo is used to recreate the decimal price in usd from the feed data.
precision is the number of decimals that the token has on dango

Rhaki (approved):

Rhaki:
Still the ibc contract needed at `dango/genesis/src/lib.rs#L416`?
```rust
// Instantiate the IBC transfer contract.
let ibc_transfer = instantiate(
    &mut msgs,
    ibc_transfer_code_hash,
    &ibc::transfer::InstantiateMsg {},
    "dango/ibc_transfer",
    "dango/ibc_transfer",
)?;
```

larry0x:
> Still the ibc contract needed at `dango/genesis/src/lib.rs#L416`?

Right now it's still used in signup flow for devnet. Will delete once we move to hyperlane

PR #435, opened 2025-01-26
feat(dango): user hyperlane metadata for new user onboarding

larry0x:


dango/types/src/account_factory/msg.rs
```diff
@@ -17,27 +18,40 @@ pub struct User {
...
+}
+
+impl Config {
+    pub fn code_hash_for(&self, ty: AccountType) -> anyhow::Result<Hash256> {
+        self.code_hashes
+            .get(&ty)
+            .copied()
+            .ok_or_else(|| anyhow!("code hash not found for account type: {ty}"))
+    }
 }
```
larry0x, line 36: I don't like this. Better if we create an abstraction over `BTreeMap` that ensures all enum variants are present in the map keys.

Issue #439, opened 2025-01-28
dango: DEX to-do list


larry0x:
- [ ] permissioned pool creation
- [ ] transient storage for incoming orders
- [ ] batch update orders method
- [ ] instant swap (our equivalent of market order)
- [ ] passive liquidity pool - xyk
- [ ] passive liquidity pool - solidly
- [ ] passive liquidity pool - stableswap

PR #447, opened 2025-01-30
feat(dango): use a transient storage for incoming orders

larry0x:
see #439


dango/dex/src/state.rs
```diff
@@ -1,18 +1,16 @@
...
-    dango_types::dex::{Direction, OrderId, PairParams},
-    grug::{Addr, Counter, Counters, Denom, IndexedMap, Map, Udec128, Uint128, UniqueIndex},
+    dango_types::{
+        dex::{Direction, OrderId, PairParams},
+        TransientMap,
+    },
+    grug::{Addr, Counter, Denom, IndexedMap, Map, Udec128, Uint128, UniqueIndex},
 };
 
+pub(crate) static INCOMING_ORDERS: TransientMap<OrderId, (OrderKey, Order)> = TransientMap::new();
```
larry0x, line 9: Noticed an issue: this static variable is shared by all threads. Meaning if one simulates a tx using the query thread, a new order would be inserted into the map, impacting the block execution thread. A solution may be to use the `thread_local!` macro.

PR #448, opened 2025-01-30
Add GraphQL subscriptions

penso:
- Manually ensure Dango can be started with httpd enabled, and GraphQL subscription works
- Add a `dango` httpd binary to start the server
- Add a `block` and `transfers` GraphQL subscriptions
- Add pubsub trait, with an in-memory implementation


dango/cli/src/start.rs
```diff
@@ -74,9 +75,12 @@ impl StartCmd {
...
-            .await?;
+    async fn run_httpd_server(context: Context) -> anyhow::Result<()> {
+        if let Err(err) =
+            indexer_httpd::server::run_server(None, None, context, config_app, build_schema).await
+        {
+            tracing::error!("Failed to run HTTP server: {:?}", err);
+        }
 
         Ok(())
     }
```
larry0x, line 86: @penso It looks like now we only prints the error but don't abort. This function always returns ok. Is this intended?

dango/cli/src/start.rs
```diff
@@ -115,6 +119,7 @@ impl StartCmd {
             codes.taxman.to_bytes().hash256(),
             codes.vesting.to_bytes().hash256(),
             codes.warp.to_bytes().hash256(),
+            codes.dex.to_bytes().hash256(),
```
larry0x, line 122: I already `dex` above (also, contracts ordered alphabetically)

grug/testing/tests/httpd.rs
```diff
@@ -238,3 +248,88 @@ async fn graphql_returns_messages() -> anyhow::Result<()> {
...
+        while rx.recv().await.is_some() {
+            let to = accounts["owner"].address;
+
+            suite
+                .send_message_with_gas(
+                    &mut accounts["sender"],
+                    2000,
+                    grug_types::Message::transfer(
+                        to,
+                        Coins::one(Denom::from_str("ugrug")?, 2_000).unwrap(),
```
larry0x, line 286: We're mixing up error handling in tests. In some places we use `unwrap` while elsewhere we use `?`.

indexer/sql/src/pubsub.rs
```diff
@@ -0,0 +1,188 @@
...
+        let pubsub = PostgresPubSub::new(pool.clone());
+        let pubsub_clone = pubsub.clone();
+
+        {
+            tokio::task::spawn(async move {
+                for idx in 1..10 {
+                    pubsub.publish_block_minted(idx).await?;
+                    tokio::time::sleep(Duration::from_secs(1)).await;
+                }
+                anyhow::Ok(())
```
larry0x, line 129: Another inconsistency: in some places we do `anyhow::Ok` while elsewhere we do `Ok::<_, anyhow::Error>`

dango/cli/src/start.rs
penso: It is intended, my biggest fear would be the node not to start because of some side issues like the local indexer http so I'll try to make it loose and not prevent starting.

PR #463, opened 2025-02-12
Add incoming orders map

pacmanifold:


larry0x:
> @larry0x I got two errors in the CI test job that I did not manage replicate locally. "Resolved" themselves when I re-ran the job. Have you had this before?

What is the error? I may have had it before. Was related to indexer

dango/dex/src/execute.rs
```diff
@@ -162,8 +163,25 @@ fn cancel_orders(ctx: MutableCtx, order_ids: BTreeSet<OrderId>) -> anyhow::Resul
...
+        // First try to load from resting orders, then from incoming orders storage.
+        // If found in neither or both through an error
+        let (((base_denom, quote_denom), direction, price, _), order) = match (
+            ORDERS.idx.order_id.may_load(ctx.storage, order_id)?,
+            INCOMING_ORDERS.may_load(ctx.storage, order_id)?,
+        ) {
+            (Some(x), None) => {
+                ORDERS.remove(ctx.storage, x.0.clone())?;
+                x
+            },
```
larry0x, line 175: @pacmanifold Don't name variables like `x`. It's hard to read. I also really hate syntax like `x.0`.

Do something like this instead:

```rust
(Some((order_key, order)), None) => {
    ORDERS.remove(ctx.storage, order_key.clone())?;
    (order_key, order)
}
```

dango/dex/src/execute.rs
```diff
@@ -162,8 +163,25 @@ fn cancel_orders(ctx: MutableCtx, order_ids: BTreeSet<OrderId>) -> anyhow::Resul
...
+            (None, Some(x)) => {
+                INCOMING_ORDERS.remove(ctx.storage, order_id);
+                x
+            },
+            (None, None) => return Err(anyhow!("order with id {order_id} not found")),
+            (Some(_), Some(_)) => {
+                return Err(anyhow!("order with id {order_id} found in both maps"))
+            },
```
larry0x, line 183: Use the `bail!` macro instead of `return Err(anyhow!)`

Also can you add backticks around order id, for better readability:

```diff
- order with id {order_id} not found
+ order with id `{order_id}` not found
```

dango/dex/src/execute.rs
```diff
@@ -209,14 +222,16 @@ pub fn cron_execute(ctx: SudoCtx) -> StdResult<Response> {
...
+    // Add all incoming orders to the orders map and clear the incoming orders map.
+    let mut pairs = BTreeSet::new();
+    let incoming_orders = INCOMING_ORDERS
+        .range(ctx.storage, None, None, IterationOrder::Ascending)
         .collect::<StdResult<Vec<_>>>()?;
+    for (_, (order_key, order)) in incoming_orders {
+        ORDERS.save(ctx.storage, order_key.clone(), &order)?;
+        pairs.insert(order_key.0);
+    }
+    INCOMING_ORDERS.clear(ctx.storage, None, None);
```
larry0x, line 234: I think this part of the code is quite ugly.

First, let's don't use `order_key.0`. Do something like:

```rust
for (_, ((pair, direction, price, order_id), order)) in incoming_orders {
    ORDERS.save(ctx.storage, (pair.clone(), direction, price, order_id), &order)?;
    pairs.insert(pair);
}
```

Sure it's more verbose, but a lot more readable imo.

Second, we have a `Map::take` method that does `Map::load` and `Map::remove` in one go. We should add a similar method `Map::drain` that does `Map::range` and `Map::clear` in one go.

Third, instead of pushing items into `pairs`, it's better to do something like:

```rust
let pairs = incoming_orders.keys().map(/* ... */).collect();
```

dango/testing/tests/dex.rs
```diff
@@ -348,3 +348,127 @@ fn dex_works(
...
+            &mut accounts.user1,
+            contracts.dex,
+            &dex::ExecuteMsg::SubmitOrder {
+                base_denom: DANGO_DENOM.clone(),
+                quote_denom: USDC_DENOM.clone(),
+                direction: Direction::Bid,
+                amount: Uint128::new(100),
+                price: Udec128::new(1),
+            },
+            Coins::one(USDC_DENOM.clone(), 100).unwrap(),
```
larry0x, line 371: Use a `grug::coins!` macro. It's easier

dango/testing/tests/dex.rs
```diff
@@ -348,3 +348,127 @@ fn dex_works(
...
+
+    // Check that the user balance has not changed
+    assert_eq!(
+        suite
+            .balances()
+            .changes(accounts.user1.address())
+            .get(&USDC_DENOM)
+            .unwrap(),
+        &BalanceChange::Unchanged
+    );
```
larry0x, line 395: We have a better syntax provided by the `grug::ResultExt` trait:

```rust
suite
    .balances()
    .changes(accounts.user1.address())
    .get(&USDC_DENOM)
    .should_equal(BalanceChange::Unchanged);
```

dango/testing/tests/dex.rs
```diff
@@ -348,3 +348,127 @@ fn dex_works(
...
+
+    // Check that order does not exist
+    let orders = suite
+        .query_wasm_smart(contracts.dex, QueryOrdersRequest {
+            start_after: None,
+            limit: None,
+        })
+        .unwrap();
+    assert!(orders.is_empty())
```
larry0x, line 404: Similarly, use the `grug::ResultExt` syntax:

```rust
suite
    .query_wasm_smart(contracts.dex, QueryOrdersRequest {
        start_after: None,
        limit: None,
    })
    .should(BTreeMap::is_empty);
```

dango/dex/src/execute.rs
larry0x: Actually you don't even need to collect `pairs` into a btreeset. You can simply do:

```rust
for (pairs, ..) in incoming_orders.keys() {
    // order clearing logics for each pair...
}
```

larry0x: See: https://github.com/left-curve/left-curve/compare/main...drain

dango/testing/tests/dex.rs
pacmanifold: the suggested syntax does not seem to exist. Changed to `suite.balances().should_change(...)`, I assume that is what you meant

dango/testing/tests/dex.rs
pacmanifold: ack

PR #468, opened 2025-02-14
feat: lending tokens and limit orders as collateral

apollo-sturdy:


dango/account/margin/src/core.rs
```diff
@@ -0,0 +1,164 @@
...
+        total_adjusted_collateral_value.checked_add_assign(adjusted_value)?;
+        limit_order_collaterals.insert(deposited)?;
+        limit_order_outputs.insert(returned)?;
+    }
+
+    // Calculate the utilization rate.
+    let utilization_rate = if total_debt_value.is_zero() {
+        // The account has no debt. Utilization is zero in this case,
+        // regardless of collateral value.
+        Udec128::ZERO
```
larry0x: Irrelevant to this PR, but I think we can make an optimization: if debt value is zero, then we can already return without querying the collaterals and limit orders?

dango/oracle/src/query.rs
```diff
@@ -38,9 +38,8 @@ pub fn query(ctx: ImmutableCtx, msg: QueryMsg) -> anyhow::Result<Json> {
 }
 
 fn query_price(ctx: ImmutableCtx, denom: Denom) -> anyhow::Result<PrecisionedPrice> {
-    PRICE_SOURCES
-        .load(ctx.storage, &denom)?
-        .get_price(ctx.storage)
+    ctx.querier
+        .query_price(ctx.contract.address(), &denom, None)
```
larry0x, line 42: Not a fan of this, but we can fix this later.

dango/oracle/src/query.rs
```diff
@@ -56,7 +55,9 @@ fn query_prices(
         .take(limit)
         .map(|res| {
             let (denom, price_source) = res?;
-            let price = price_source.get_price(ctx.storage)?;
+            let price =
+                ctx.querier
+                    .query_price(ctx.contract.address(), &denom, Some(price_source))?;
```
larry0x, line 60: And this

dango/types/src/oracle/price_source.rs
```diff
@@ -27,24 +27,6 @@ pub enum PriceSource {
...
-                Ok(price.with_precision(*precision))
-            },
-            Self::Pyth { id, precision } => {
-                let price = PRICES.load(storage, *id)?;
-                Ok(price.with_precision(*precision))
-            },
-        }
-    }
+    /// A price source for an LP token of the lending pool.
+    LendingLP {},
```
larry0x: If you have to put abbreviations such as "LP" in enums, do `LendingLp` instead of `LendingLP`. Otherwise in JSON this gets serialized as `"lending_l_p"` which isn't ideal.

dango/account/margin/src/core.rs
```diff
@@ -0,0 +1,183 @@
...
+            querier.query_price(app_cfg.addresses.oracle, &returned.denom, None)?;
+        let deposited_value = deposited_price.value_of_unit_amount(deposited.amount)?;
+        let returned_value = returned_price.value_of_unit_amount(returned.amount)?;
+
+        let min_value = min(deposited_value, returned_value);
+
+        let collateral_power = app_cfg
+            .collateral_powers
+            .get(&deposited.denom)
+            .ok_or_else(|| anyhow!("collateral power for denom {} not found", deposited.denom))?;
```
larry0x, line 149: Collateral power should just be zero if not found. Imagine someone launches a random shitcoin. Of course it won't have a collateral power. But people may still trade it on the DEX.

apollo-sturdy: True. Will fix

dango/account/margin/src/core.rs
```diff
@@ -0,0 +1,164 @@
...
+
+        let collateral_power =
+            app_cfg
+                .collateral_powers
+                .get(&deposited.denom)
+                .ok_or(anyhow::anyhow!(
+                    "collateral power for denom {} not found",
+                    deposited.denom
+                ))?;
+        let adjusted_value = min_value.checked_mul(collateral_power.clone().into_inner())?;
```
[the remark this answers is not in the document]
apollo-sturdy: True, will change it.

dango/account/margin/src/execute.rs
```diff
@@ -217,11 +222,21 @@ pub fn liquidate(ctx: MutableCtx, liquidation_denom: Denom) -> anyhow::Result<Re
         repay_coins.clone(),
     )?;
 
+    // Create message to cancel all the user's limit orders
+    let cancel_msg = Message::execute(
+        app_cfg.addresses.dex,
+        &dango_types::dex::ExecuteMsg::CancelOrders {
+            order_ids: btree_set! {},
```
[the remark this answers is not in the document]
apollo-sturdy: Sure. Can implement the enum

dango/account/margin/src/margin_querier.rs
```diff
@@ -86,15 +92,68 @@ where
...
+                        deposited.denom
+                    ))?;
+            let adjusted_value = min_value.checked_mul(collateral_power.clone().into_inner())?;
+
+            total_collateral_value.checked_add_assign(min_value)?;
+            total_adjusted_collateral_value.checked_add_assign(adjusted_value)?;
+            limit_order_collaterals.insert(deposited)?;
+            limit_order_outputs.insert(returned)?;
+        }
+
```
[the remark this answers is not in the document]
apollo-sturdy: Oops, yes should be deleted.

dango/account/margin/src/execute.rs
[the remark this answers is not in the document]
apollo-sturdy: fixed

dango/account/margin/src/core.rs
apollo-sturdy: Mm but would need some refactoring, since we also want to still be able to query full HealthResponse via QueryMsg.

dango/account/margin/src/core.rs
apollo-sturdy: fixed

dango/account/margin/src/core.rs
```diff
@@ -0,0 +1,188 @@
...
+            ),
+        };
+
+        let offer_price = querier.query_price(app_cfg.addresses.oracle, &offer.denom, None)?;
+        let offer_value = offer_price.value_of_unit_amount(offer.amount)?;
+        let offer_collateral_power = app_cfg
+            .collateral_powers
+            .get(&offer.denom)
+            .map(|x| x.clone().into_inner())
+            .unwrap_or_else(|| Udec128::ZERO);
```
larry0x, line 142: `.unwrap_or` is ok here because `Udec128::ZERO` is a constant

dango/dex/src/query.rs
```diff
@@ -142,10 +142,34 @@ fn query_orders_by_pair(
...
-    _limit: Option<u32>,
+    ctx: ImmutableCtx,
+    user: Addr,
+    start_after: Option<OrderId>,
+    limit: Option<u32>,
 ) -> StdResult<BTreeMap<OrderId, OrdersByUserResponse>> {
-    todo!();
+    let start = start_after.map(|order_id| {
+        let (((base_denom, quote_denom), direction, price, _), _) =
+            ORDERS.idx.order_id.load(ctx.storage, order_id).unwrap();
```
larry0x, line 152: We can't unwrap here, otherwise an attack can query an non-existing order id, then this would panic and crash the node. We need to avoid all sorts of panics in `RustVm` contracts.

dango/dex/src/execute.rs
```diff
@@ -158,38 +158,46 @@ fn submit_order(
...
+        OrderIds::All => ORDERS
+            .idx
+            .user
+            .prefix(ctx.sender)
+            .range(ctx.storage, None, None, IterationOrder::Ascending)
+            .map(|order| Ok((order?, false)))
+            .chain(
+                INCOMING_ORDERS
+                    .prefix(ctx.sender)
+                    .range(ctx.storage, None, None, IterationOrder::Ascending)
```
larry0x, line 175: Instead of `.range` you can use `.values` here

dango/dex/src/query.rs
apollo-sturdy: Good point

PR #472, opened 2025-02-20
Add more optimized pubsub

penso:
Add a more optimized psql pubsub using a single SQL connections whatever the amount of subscribers there is.


indexer/sql/src/pubsub.rs
```diff
@@ -112,16 +25,19 @@ pub enum PubSubType {
...
+        sqlx::postgres::{PgListener, PgPoolOptions},
+        std::time::Duration,
         tokio_stream::StreamExt,
     };
 
     #[ignore]
     #[tokio::test]
     async fn test_postgres_pubsub() -> anyhow::Result<()> {
-        let pool = sqlx::PgPool::connect("postgres://postgres@postgres/grug_test").await?;
+        let pool = sqlx::PgPool::connect("postgres://postgres@localhost/grug_test").await?;
```
penso, line 38: Use ENV `DB_HOST` to allow automatic docker containers in the CI, but `localhost` when running locally on computers.

PR #474, opened 2025-02-21
feat(dango): provide withdraw liquidity

pacmanifold:


dango/dex/src/execute.rs
```diff
@@ -221,6 +298,128 @@ fn cancel_orders(ctx: MutableCtx, order_ids: BTreeSet<OrderId>) -> anyhow::Resul
...
+    let lp_mint_amount = lp_supply.checked_mul_dec_floor(mint_ratio)?;
+
+    // Apply swap fee to unbalanced provision
+    fn abs_diff(a: Uint128, b: Uint128) -> Uint128 {
+        if a > b {
+            a - b
+        } else {
+            b - a
+        }
+    }
```
pacmanifold, line 344: @larry0x should we implement this on `Int` instead? or as part of the `Number` trait?

dango/testing/tests/dex.rs
```diff
@@ -466,3 +466,296 @@ fn submit_and_cancel_order_in_same_block() {
...
+        .into_iter()
+        .map(|Coin { denom, amount }| {
+            (denom.clone(), BalanceChange::Increased(amount.into_inner()))
+        })
+        .collect();
+    changes.extend(decreases.into_iter().map(|Coin { denom, amount }| {
+        (denom.clone(), BalanceChange::Decreased(amount.into_inner()))
+    }));
+    changes
+}
```
pacmanifold, line 757: should I put this in some place with shared test helpers? Or should I impl it on `BTreeMap<Denom, BalanceChange>`?

dango/dex/src/execute.rs
```diff
@@ -238,6 +315,125 @@ fn cancel_orders(ctx: MutableCtx, order_ids: OrderIds) -> anyhow::Result<Respons
...
+    let fee_rate = Udec128::checked_from_ratio(
+        abs_diff(a, avg_reserves).checked_add(abs_diff(b, avg_reserves))?,
+        sum_reserves,
+    )?
+    .checked_mul(
+        pool.swap_fee
+            .checked_div(Udec128::checked_from_ratio(2, 1)?)?,
+    )?;
+    let lp_mint_amount_after_fees =
+        lp_mint_amount.checked_mul_dec_floor(Udec128::ONE.checked_sub(fee_rate)?)?;
```
larry0x: I think these logic should be in the `PassiveLiquidityPool` trait. Let's keep `execute.rs` as lean as possible.

dango/types/src/dex.rs
```diff
@@ -1,8 +1,32 @@
...
+        sync::LazyLock,
+    },
 };
 
+/// The namespace used for dex.
+///
+/// E.g.,
+///
+/// - `dex/eth`
+/// - `dex/usdc`
```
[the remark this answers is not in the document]
patriklaurell: absolutely unnecessary

PR #489, opened 2025-02-28
Add more subscription types

penso:


indexer/httpd/src/graphql/types/transaction.rs
```diff
@@ -23,12 +22,30 @@ pub struct Transaction {
...
+    Cron,
+    Tx,
+}
+
+impl From<i16> for Category {
+    fn from(status: i16) -> Category {
+        match status {
+            0 => Category::Cron,
+            1 => Category::Tx,
+            _ => panic!("Invalid transaction category"),
```
larry0x, line 38: Is it safe to panic here? If a panic will it kill the entire node process?

larry0x: Will the `i16` be input by the user? Because if yes, then a user can send an invalid `i16` and cause the node to panic and halt...?

PR #481, opened 2025-02-24
feat(dango): impl `BatchSubmitOrders`

pacmanifold:

PR #492, opened 2025-02-28
feat(dango/dex): add `batch_update_orders` method

### LINKED ISSUE #491: dango: batch update orders
add a single method to submit or cancel multiple orders in one batch. replaces `submit_order` and `cancel_orders`

```rust
enum ExecuteMsg {
    BatchUpdateOrders {
        submits: Vec<Submit>,
        cancels: OrderIds,
    }
}

struct Submit {
    base_denom: Denom,
    quote_denom: Denom,
    direction: Direction,
    amount: Uint128,
    price: Udec128,
}
```

the logic of the function would be:

```rust
fn batch_update_orders(
    ctx: MutableCtx,
    submits: Vec<Submit>,
    cancels: OrderIds,
) -> anyhow::Result<Response> {
    // the coins that should be refunded to the user due to canceling orders
    let mut refunds = Coins::new();

    for order_id in order_ids {
        // 1. cancel the order
        // 2. increment `refunds`
    }

    // the coins that the user is expected to deposit in order to submitting orders
    let mut deposits = Coins::new();

    for submit in submits {
        // 1. submit the order
        // 2. increment `deposits`
    }

    // compute how much coins should be sent back to the user.
    //
    // this should be whatever user sent to the contract, plus the refunds from
    // cancelled orders, minus the expected deposits for submitting orders.
    //
    // if the user didn't send enough funds, this here should error.
    let output = ctx
        .funds
        .add_many(refunds)?
        .deduct_many(deposits)?;

    Ok(Response::new()
        .add_message(Message::transfer(bank, ctx.sender, output)?))
}
```

larry0x:
closes #491 

replaces #481

### DIFF
--- a/dango/dex/src/execute.rs
+++ b/dango/dex/src/execute.rs
@@ -7,9 +7,9 @@ use {
     dango_types::{
         bank,
         dex::{
-            Direction, ExecuteMsg, InstantiateMsg, OrderCanceled, OrderFilled, OrderIds,
-            OrderSubmission, OrderSubmitted, OrdersMatched, PairUpdate, PairUpdated, LP_NAMESPACE,
-            NAMESPACE,
+            CreateLimitOrderRequest, Direction, ExecuteMsg, InstantiateMsg, OrderCanceled,
+            OrderFilled, OrderIds, OrderSubmitted, OrdersMatched, PairUpdate, PairUpdated,
+            LP_NAMESPACE, NAMESPACE,
         },
     },
     grug::{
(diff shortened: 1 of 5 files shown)


dango/dex/src/execute.rs
```diff
@@ -84,6 +80,165 @@ fn batch_update_pairs(ctx: MutableCtx, updates: Vec<PairUpdate>) -> anyhow::Resu
...
+    // --------------------------- 1. Cancel orders ----------------------------
+
+    // First, collect all orders to be cancelled into memory.
+    let orders = match cancels {
+        // Cancel all orders.
+        Some(OrderIds::All) => ORDERS
+            .idx
+            .user
+            .prefix(ctx.sender)
+            .range(ctx.storage, None, None, IterationOrder::Ascending)
```
pacmanifold, line 102: could we ever conceivably run out of gas here? i.e. do we need to have pagination ability? would that be a problem in liquidation?

pacmanifold (commented):
lgtm! just one question. Will put new swap PR on top of this and add the missing test

dango/dex/src/execute.rs
larry0x: actually there's gas consumption when reading data from DB, but computations in the VM don't have gas consumption yet.

PR #499, opened 2025-03-03
feature: Hyperlane agent changes

Rhaki:


hyperlane/va/src/execute.rs
```diff
@@ -66,8 +66,9 @@ fn announce(
         ctx.storage,
         validator,
         |maybe_storage_locations| -> StdResult<_> {
-            let mut storage_locations = maybe_storage_locations.unwrap_or_default();
-            storage_locations.insert(storage_location.clone());
+            let mut storage_locations = maybe_storage_locations.unwrap_or(UniqueVec::new(vec![])?);
```
larry0x, line 69: can use `UniqueVec::new_unchecked` here

larry0x: or just `.unwrap_or_default`

PR #503, opened 2025-03-04
feat(dango): implement `Swap` for XYK pool

pacmanifold:


grug/types/src/coin_pair.rs
```diff
@@ -160,6 +160,58 @@ impl CoinPair {
...
+    }
+
+    /// Return the amount of the given denom in the coin pair.
+    pub fn amount_of(&self, denom: &Denom) -> Uint128 {
+        if self.first().denom == denom {
+            *self.first().amount
+        } else {
+            *self.second().amount
+        }
+    }
```
larry0x, line 176: i think you need to account for the case where `denom` equals neither the two denoms in the `CoinPair`

pacmanifold: good catch. will return a result instead

pacmanifold:
I realised there is an issue with this implementation when trying to implement the routing. After rebasing there is no way to place a sell or buy order for `x` amount of USDC if the pair is saved as `{ base_denom: ETH, quote_denom: USDC }`. I.e. there is no way to specify a swap amount in terms of the quote denom. I will put a proposed fix up today

grug/types/src/coin_pair.rs
pacmanifold: fixed

pacmanifold:
closing in favor of #625

PR #518, opened 2025-03-08
refactor(hyperlane): delete hooks, merge merkle tree functionality into mailbox

larry0x:
closes: #516


larry0x:
This PR removes "hooks" from the `hyperlane-mailbox` contract.

For context, the Hyperlane protocol defines two hook contracts:

- the `required_hook`, which is typically set to the `InterchainGasPaymaster` (IGP) contract, which handles fees;
- the `default_hook`, which is typically set to the Merkle tree hook.

We don't use the IGP contract. IGP always requires fees in the chain's gas currency (e.g. on Ethereum, it would be ETH). In contrary, we want an experience similar to centralized exchanges: the fee is always paid in the asset you're withdrawing. E.g. withrawing 1.0 BTC, we deduct a fee of 0.0001 BTC, user gets 0.9999 BTC. The way IGP works obviously doesn't fit for this purpose. Therefore, we implemented a separate fee mechanism in the Warp contract.

Regarding the `default_hook`, in practice, one should _always_ use the Merkle tree hook. Hyperlane's relayer software only relays messages that have been included in the Merkle tree. It would be much simpler that Merkle tree logics are simply included as part of the mailbox contract, instead of being a separate and optional contract.

PR #523, opened 2025-03-10
Hyperlane/audit fix

Zexsor33:
Apply suggestion to hyperlane contract from cw-hyperlane audit:
- adding fee for validator announce;
- check in mailbox if destination domain is different from local domain


dango/genesis/src/lib.rs
```diff
@@ -369,7 +370,10 @@ where
     let va = instantiate(
         &mut msgs,
         hyperlane_va_code_hash,
-        &va::InstantiateMsg { mailbox },
+        &va::InstantiateMsg {
+            mailbox,
+            announce_fee_per_byte: Coin::new(DANGO_DENOM.clone(), 100).unwrap(),
```
larry0x, line 375: we should add this to `GenesisConfig`

hyperlane/mailbox/src/execute.rs
```diff
@@ -46,6 +46,12 @@ fn dispatch(
     let cfg = CONFIG.load(ctx.storage)?;
     let (nonce, _) = NONCE.increment(ctx.storage)?;
 
+    // Ensure the destination domain is not the local domain.
+    ensure!(
+        destination_domain != cfg.local_domain,
+        "Destination domain is the same as the local domain"
```
larry0x, line 52: do not uppercase the error message. see rust API guidelines:

https://github.com/rust-lang/api-guidelines/blob/master/src/interoperability.md#examples-of-error-messages

hyperlane/types/src/va.rs
```diff
@@ -37,6 +39,9 @@ pub enum QueryMsg {
     AnnouncedStorageLocations {
         validators: BTreeSet<HexByteArray<20>>,
     },
+    /// Query the cost of announcing a storage location.
+    #[returns(Coin)]
+    CalculateAnnounceFee { storage_location: String },
```
larry0x, line 44: query function name should be a noun phrase, not a verb, to be consistent with all the other methods above. so just `AnnounceFee`, not `CalculateAnnounceFee`

Issue #521, opened 2025-03-09
issues


larry0x:
1. The indexer DB can't be in `~/.dango`, because if so it'd impossible to run multiple versions of the DB (e.g. for multiple git branches).
2. It may be better to put the `app.toml` in the repo's root, for the same reason--to run multiple versions of the code with different configurations.
3. The config-parser code is better extracted to a standalone crate, so it can be used not only in dango-cli but elsewhere as well.
4. Rename `postgres_url` back to `database_url`, since the DB isn't necessarily postgres (mysql and sqlite are also supported).

PR #525, opened 2025-03-11
localnet

larry0x:
closes: #521


networks/localdango/configs/dango/config/app.toml
```diff
@@ -52,10 +52,10 @@ port = 8080
 [tendermint]
 
 # Tendermint RPC listening address.
-rpc_addr = "http://localhost:26657"
+rpc_addr = "http://0.0.0.0:26657"
 
 # Tendermint ABCI listening address.
-abci_addr = "http://localhost:26658"
+abci_addr = "0.0.0.0:26658"
```
larry0x, line 58: dango and cometbft are running in different containers, so listening address won't be `localhost`.

Ideally we have dango only listen to the cometbft container, but putting

```toml
abci_addr = "cometbft:26658"
```

doesn't work. I don't know how to make it work.

larry0x: the error is:

```
Error: failed to start ABCI server: Os { code: 99, kind: AddrNotAvailable, message: "Cannot assign requested address" }
```

it seems this isn't related to DNS resolution. the hostname `cometbft` is properly resolved to an IP (`192.128.*.*`) using `tokio::net::lookup_host`.

rather, according to tokio docs, it's because of "the requested address was not local":

https://doc.rust-lang.org/stable/std/io/enum.ErrorKind.html#variant.AddrNotAvailable

larry0x: @penso i tried various things and none worked.

the TCP protocol doesn't natively have a filter for connections. meaning we can only either listen on `127.0.0.1:{port}` or `0.0.0.0:{port}`. we can't say "only listen for connections from IP `xxx:xxx:xxx:xxx`".

i've also tried on the docker compose side, such as setting up an internal network. but nothing seems to work.

i think this is what it is. we have to have dango listen on `0.0.0.0:26658`, and simply trust the host closes port 26658 from the external world e.g. via a firewall.

i think this is fine. if you run `dango` as a binary without using docker, you still need to prevent anyone else other than localhost from accessing 26658. so there's really no difference.

penso (approved):

PR #527, opened 2025-03-11
Refactor suggestions for PythClient

penso:


pyth/client/src/middleware_cache.rs
```diff
@@ -1,202 +1,171 @@
...
 
                 // Load the data from disk.
                 let loaded_vaas = cache_file.load::<Vec<Vec<Binary>>>().unwrap();
-                self.stored_vaas
-                    .insert(filename.clone(), loaded_vaas.into_iter());
+                stored_vaas.insert(filename, loaded_vaas.into_iter());
             }
         }
+
+        stored_vaas
```
penso, line 83: Good place to change the published time if we need multiple different values from Pyth.

pyth/client/src/middleware_cache.rs
```diff
@@ -1,202 +1,171 @@
...
 
-            // Sleep for 0,5 seconds.
-            sleep(Duration::from_millis(500));
-        }
+        workspace_root.join("pyth/client/testdata").join(id)
     }
+}
+
+#[async_trait]
+impl PythClientTrait for PythMiddlewareCache {
```
penso, line 117: rename to `PythCacheClient`.

pyth/client/src/middleware_cache.rs
```diff
@@ -1,202 +1,171 @@
...
+        I: IntoIterator + Lengthy + Send + Clone,
         I::Item: ToString,
-        T: ToString,
     {
-        self.load_or_retrieve_data(ids.clone(), base_url);
+        let stored_vaas = self.get_latest_vaas(ids.clone()).unwrap();
+
+        let stream = stream! {
+            loop {
+                yield stored_vaas.clone();
```
penso, line 132: Could use this to change the vaas published time.

pyth/client/tests/common_function.rs
```diff
@@ -170,3 +176,54 @@ where
...
+    let client = PythClient::new(PYTH_URL);
+
+    let mut stream = client
+        .stream(NonEmpty::new_unchecked(ids1.clone()))
+        .await
+        .unwrap();
+
+    for _ in 0..5 {
+        let Some(vaas) = stream.next().await else {
+            continue;
```
penso, line 203: this could loop 5 times with none, to be fixed.

dango/testing/tests/oracle.rs
```diff
@@ -141,7 +141,7 @@ fn oracle() {
 fn double_vaas() {
     let (mut suite, mut accounts, oracle) = setup_oracle_test();
 
-    let mut pyth_client = PythClient::new("not_real_url").with_middleware_cache();
+    let pyth_client = PythMiddlewareCache::new("not_real_url");
```
penso, line 144: Use Into<Url> instead of String.

pyth/client/src/middleware_cache.rs
Zexsor33: Actually we can't. The problem is that if we change publish_time inside the vaas, the verification inside the Oracle contract would fail because the signature is not the same (PP test). To solve this problem, in my PR I stored an `Iterator` of 15 vaas and every time the stream needs to return a value, it calls the `next()` method and returns a newer value

pyth/client/src/client_cache.rs
```diff
@@ -0,0 +1,189 @@
...
+                    ty: "cache",
+                    key: filename.to_string_lossy().to_string(),
+                });
+            }
+        }
+
+        self.vaas_index.fetch_add(1, Ordering::SeqCst);
+
+        Ok(return_vaas)
+    }
```
penso, line 186: @Zexsor33 This feels overly complex for no reason, you could just use the same pattern as the `stream` method:

```rust
    fn get_latest_vaas<I>(&self, ids: NonEmpty<I>) -> Result<Vec<Binary>, Self::Error>
    where
        I: IntoIterator + Clone + Lengthy,
        I::Item: ToString,
    {
        // Load the data.
        let stored_vaas = Self::load_or_retrieve_data(self.base_url.clone(), ids.clone());
        let index = self.vaas_index.load(Ordering::Acquire) as usize;

        let result = stored_vaas
            .iter()
            .filter_map(|(_, v)| v.get(index).cloned())
            .flatten()
            .collect::<Vec<_>>();

        self.vaas_index.fetch_add(1, Ordering::SeqCst);

        Ok(result)
    }
```

penso: And then you can remove the `DataNotFound` error enum from `error::Error`

pyth/client/src/error.rs
```diff
@@ -0,0 +1,15 @@
...
+    Reqwest(#[from] reqwest::Error),
+
+    #[error(transparent)]
+    CannotClone(#[from] reqwest_eventsource::CannotCloneRequestError),
+
+    #[error("data not found! type: {ty}, storage key: {key}")]
+    DataNotFound { ty: &'static str, key: String },
+
+    #[error(transparent)]
+    ParseError(#[from] ParseError),
```
penso, line 14: This gives `warning: variant name ends with the enum's name`, should be renamed `CannotParseURL` for example.

pyth/client/src/client_cache.rs
```diff
@@ -0,0 +1,189 @@
...
+    pub fn new<U: IntoUrl>(base_url: U) -> Result<Self, error::Error> {
+        Ok(Self {
+            base_url: base_url.into_url()?,
+            vaas_index: Arc::new(AtomicU64::new(0)),
+        })
+    }
+
+    /// Load data from cache or retrieve it from the source.
+    pub fn load_or_retrieve_data<I>(
+        base_url: Url,
```
penso, line 40: Since we're cloning `base_url` inside this method (because of the loop) you can change that to `&Url` and avoid base_url clones in the methods calling `load_or_retrieve_data`

pyth/client/src/client_cache.rs
```diff
@@ -0,0 +1,189 @@
...
+                            if let Some(vaas) = stream.next().await {
+                                values.push(vaas);
+                            }
+                        }
+
+                        values
+                    });
+
+                    // Store the data in the cache.
+                    Self::store_data(element, values).unwrap();
```
penso, line 81: We already have `cache_file` so we don't need `Self::store_data()`, we could have a better version of `load_or_retrieve_data`:

```rust
    /// Load data from cache or retrieve it from the source.
    pub fn load_or_retrieve_data<I>(
        base_url: &Url,
        ids: NonEmpty<I>,
    ) -> HashMap<PathBuf, Vec<Vec<Binary>>>
    where
        I: IntoIterator + Lengthy + Clone,
        I::Item: ToString,
    {
        let mut stored_vaas = HashMap::new();

        // Load data for each id.
        for id in ids.into_inner() {
            let element = id.to_string();
            let filename = Self::cache_filename(&element);

            // If the file is not in memory, try to read from disk.
            stored_vaas.entry(filename.clone()).or_insert_with(|| {
                let cache_file = DiskPersistence::new(filename.clone(), true);

                if cache_file.exists() {
                    return cache_file.load::<Vec<Vec<Binary>>>().unwrap();
                }

                // If the file does not exists, retrieve the data from the source.
                let rt = Runtime::new().unwrap();
                let values = rt.block_on(async {
                    let client = PythClient::new(base_url.clone()).unwrap();

                    let mut stream = client
                        .stream(NonEmpty::new(vec![id.to_string()]).unwrap())
                        .await
                        .unwrap();

                    let mut values = vec![];
                    while values.len() < 15 {
                        if let Some(vaas) = stream.next().await {
                            values.push(vaas);
                        }
                    }

                    values
                });

                // Store the data in the cache.
                cache_file.save(&values).unwrap();
                values
            });
        }

        stored_vaas
    }
```

pyth/client/src/client_cache.rs
```diff
@@ -0,0 +1,189 @@
...
+                    let values = rt.block_on(async {
+                        let client = PythClient::new(base_url.clone()).unwrap();
+
+                        let mut stream = client
+                            .stream(NonEmpty::new(vec![id.to_string()]).unwrap())
+                            .await
+                            .unwrap();
+
+                        let mut values = vec![];
+                        while values.len() < 15 {
```
penso, line 71: Maybe we should put that `15` in a static at the top of the file to easily change that if needed.

dango/app/src/proposal_preparer.rs
```diff
@@ -57,10 +78,31 @@ impl grug_app::ProposalPreparer for ProposalPreparer {
...
+                    if let Ok(pyth_ids) = NonEmpty::new(pyth_ids) {
+                        client.connect_stream(pyth_ids);
+                    }
+                    *pyth_client = client;
+                }
+            },
+            Err(_) => {
+                error!("Failed to update Pyth ids");
+            },
+        }
```
Zexsor33, line 105: What do you think about moving this part of code inside `PythPPHandler`? We could have a function `update_stream` that checks if the ids have changed and starts a new stream if so. I think we don't need to clone the Handler or create a new one, we can just modify the instance we have, keep the same `Shared`, update `current_ids`, replace `stoppable_thread`.

penso: I think when I quickly tried I had issue to do like this, without a mutex, so why I took it out. I think it's a good idea to try yes.

larry0x:
@penso confirming this PR is ready?

PR #533, opened 2025-03-12
refractor(grug + dango): Bank events

### LINKED ISSUE #531: ensure the transfer event is emitted on all cases where balances change
i don't recall whether we emit the transfer emit on minting/burning, or force transfer

or when someone attaches `funds` in an execute message

let's check and ensure the event is emitted in all these cases, because the liquidator bot needs it to track account balances

---

how does the liquidator track user balances?

first, we should use an approach that is safe: it always works, even though it can be slower

suggestion:

- maintain a list of all margin account addresses
- listen to events:
  - if there's an "open new margin account event", update the set of margin accounts
  - go though the events and look for addresses. if an address appears in the set of margin accounts, re-query the balances

Rhaki:
closes #531

### DIFF
--- a/dango/genesis/examples/build_genesis.rs
+++ b/dango/genesis/examples/build_genesis.rs
@@ -11,7 +11,7 @@ use {
         taxman,
     },
     grug::{
-        btree_map, coins, Bounded, Coins, Denom, Duration, HashExt, Inner, Json, JsonDeExt,
+        btree_map, coins, Bounded, Coin, Coins, Denom, Duration, HashExt, Inner, Json, JsonDeExt,
         JsonSerExt, Udec128,
     },
     hex_literal::hex,
@@ -162,6 +162,7 @@ fn main() {
         wormhole_guardian_sets: GUARDIAN_SETS.clone(),
         hyperlane_local_domain: 88888888,
         hyperlane_ism_validator_sets: btree_map! {},
+        hyperlane_va_announce_fee_per_byte: Coin::new(USDC_DENOM.clone(), 100).unwrap(),
         warp_routes: btree_map! {},
     })
     .unwrap();
--- a/dango/genesis/src/lib.rs
+++ b/dango/genesis/src/lib.rs
@@ -10,9 +10,9 @@ use {
         taxman, vesting, warp,
     },
     grug::{
-        btree_map, btree_set, Addr, Binary, Coins, Config, ContractBuilder, ContractWrapper, Denom,
-        Duration, GenesisState, Hash256, HashExt, JsonSerExt, Message, Permission, Permissions,
-        ResultExt, StdResult, GENESIS_SENDER,
+        btree_map, btree_set, Addr, Binary, Coin, Coins, Config, ContractBuilder, ContractWrapper,
+        Denom, Duration, GenesisState, Hash256, HashExt, JsonSerExt, Message, Permission,
+        Permissions, ResultExt, StdResult, GENESIS_SENDER,
     },
     hyperlane_types::{
         isms::{self, multisig::ValidatorSet},
@@ -97,6 +97,8 @@ pub struct GenesisConfig<T> {
     pub hyperlane_local_domain: Domain,
     /// Hyperlane validator sets for remote domains.
     pub hyperlane_ism_validator_sets: BTreeMap<Domain, ValidatorSet>,
+    /// Hyperlane validator announce fee rate.
+    pub hyperlane_va_announce_fee_per_byte: Coin,
     /// Warp token transfer routes.
     pub warp_routes: BTreeMap<(Denom, Domain), Addr32>,
     // TODO: add margin account parameters (collateral powers and liquidation)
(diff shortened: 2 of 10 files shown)


grug/types/src/response.rs
```diff
@@ -67,25 +67,23 @@ impl Response {
...
 
-    pub fn add_subevent(mut self, event: ContractEvent) -> Self {
-        self.subevents.push(event);
-        self
-    }
-
-    pub fn may_add_subevent(mut self, maybe_event: Option<ContractEvent>) -> Self {
+    pub fn may_add_event<E>(mut self, maybe_event: Option<E>) -> StdResult<Self>
+    where
+        E: TryInto<ContractEvent, Error = StdError>,
```
larry0x, line 78: @Rhaki if i already have a `ContractEvent`, then i can't use it here as `maybe_event`, because `ContractEvent` itself doesn't implement `TryInto<ContractEvent, Err = StdError>`.

instead, `ContractEvent` itself implements `TryInto<ContractEvent, Err = Infallible>`, and `Inflallible` implements `Into<StdError>`. therefore you can make it work by changing the function signature to:

```rust
fn may_add_event<E>(mut self, maybe_event: Option<E>) -> StdResult<Self>
where
    E: TryInto<ContractError>,
    StdError: From<E::Error>,
{ /* ... */ }
```

dango/types/src/bank/events.rs
```diff
@@ -0,0 +1,17 @@
+use {
+    grug::{Addr, Denom, Uint128},
+    std::collections::BTreeMap,
+};
+
+#[grug::derive(Serde)]
+#[grug::event("balance_changes")]
+pub struct BalanceChangesEvent {
```
larry0x, line 8: can you remove the word "Event" from the struct name? to be consistent with other events in the codebase.

dango/bank/src/execute.rs
```diff
@@ -248,34 +285,104 @@ fn decrease_supply(
...
+            self.user_changes
+                .entry(event.address)
+                .or_default()
+                // There are no way to add multiple denom for the same address.
+                .insert(event.denom, (event.amount, event.direction));
         }
-    })
+
+        Ok(())
+    }
```
larry0x, line 354: this function doesn't need to return a `Result` right? it never fails

dango/bank/src/execute.rs
```diff
@@ -248,34 +285,104 @@ fn decrease_supply(
...
+}
+
+impl EventBuilder {
+    pub fn add_event(&mut self, event: Option<BalanceChange>) -> Result<(), StdError> {
+        if let Some(event) = event {
+            self.user_changes
+                .entry(event.address)
+                .or_default()
+                // There are no way to add multiple denom for the same address.
+                .insert(event.denom, (event.amount, event.direction));
```
larry0x, line 350: better to add a `debug_assert!` to ensure the return value of `insert` is `None`.

dango/bank/src/execute.rs
```diff
@@ -248,34 +285,104 @@ fn decrease_supply(
...
+        .may_modify(storage, (&address, &denom), |maybe_balance| {
+            let balance = maybe_balance.unwrap_or(Uint128::ZERO).checked_sub(amount)?;
+            // If balance is reduced to zero, delete it, to save disk space.
+            if balance.is_zero() {
+                Ok(None)
+            } else {
+                Ok(Some(balance))
+            }
+        })
+        .map(|maybe_balance_change| {
```
larry0x, line 328: `maybe_balance_change` is not the correct variable name. should be `maybe_balance`. because this variable is the balance after the decrease, not the diff/delta.

dango/bank/src/execute.rs
```diff
@@ -248,34 +285,104 @@ fn decrease_supply(
...
+            maybe_balance_change.map(|_| BalanceChange {
+                address,
+                denom,
+                amount,
+                direction: BalanceChangeDirection::Decrease,
+            })
```
larry0x, line 334: i don't understand the logic here. if balance is decreased to zero, then `maybe_balance_change` (should be correctly renamed to `maybe_balance`) would be a `None`, then you don't emit an event?

dango/bank/src/execute.rs
```diff
@@ -248,34 +285,104 @@ fn decrease_supply(
...
+        })
+}
+
+#[derive(Default)]
+struct EventBuilder {
+    user_changes: BTreeMap<Addr, BTreeMap<Denom, (Uint128, BalanceChangeDirection)>>,
+}
```
larry0x, line 341: can you extract this to a separate `event_builder.rs` file

dango/types/src/bank/events.rs
```diff
@@ -0,0 +1,17 @@
...
+    grug::{Addr, Denom, Uint128},
+    std::collections::BTreeMap,
+};
+
+#[grug::derive(Serde)]
+#[grug::event("balance_changes")]
+pub struct BalanceChanges {
+    pub address: Addr,
+    pub changes: BTreeMap<Denom, (Uint128, BalanceChangeDirection)>,
+}
```
larry0x, line 11: may be simpler to define this as:

```rust
struct BalanceChanges {
    pub address: Addr,
    pub increases: Coins,
    pub decreases: Coins,
}
```

Issue #452, opened 2025-02-05
dango: add bank events


larry0x:
Currently, only transfers done by `MsgTransfer` emits an event. However, Dango bank contract allow a few other other ways to transfer, including mint, burn, force transfer, batch transfer, recover transfer. These need events such that they can be indexed.

PR #535, opened 2025-03-12
refractor(dango + grug): Bank events

Rhaki:
closes: #452, #531

### DIFF
--- a/dango/bank/src/execute.rs
+++ b/dango/bank/src/execute.rs
@@ -1,10 +1,12 @@
 use {
     crate::{BALANCES, METADATAS, NAMESPACE_OWNERS, ORPHANED_TRANSFERS, SUPPLIES},
     anyhow::{bail, ensure},
-    dango_types::bank::{ExecuteMsg, InstantiateMsg, Metadata},
+    dango_types::bank::{
+        Burned, ExecuteMsg, InstantiateMsg, Metadata, Minted, Received, Sent, TransferOrphaned,
+    },
     grug::{
-        Addr, BankMsg, Coin, Coins, Denom, IsZero, MutableCtx, Number, NumberConst, Part,
-        QuerierExt, Response, StdError, StdResult, Storage, SudoCtx, Uint128,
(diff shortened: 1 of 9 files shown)


dango/bank/src/execute.rs
```diff
@@ -206,12 +252,39 @@ pub fn bank_execute(ctx: SudoCtx, msg: BankMsg) -> StdResult<Response> {
...
+        })?
+        .may_add_event(if !recipient_exists {
+            Some(TransferOrphaned {
+                from: msg.from,
+                to: msg.to,
+                coins: msg.coins,
+            })
+        } else {
+            None
+        })
```
larry0x, line 287: idk if there's a cleaner syntax for doing this

larry0x: we can have a better syntax if `ContractEvent::new` takes a `&U` instead of `U`. let's fix this in a future PR?

dango/bank/src/execute.rs
```diff
@@ -185,11 +219,23 @@ fn recover_transfer(ctx: MutableCtx, sender: Addr, recipient: Addr) -> anyhow::R
...
+        increase_balance(ctx.storage, &ctx.sender, coin.denom, *coin.amount)?;
     }
 
-    Ok(Response::new())
+    Ok(Response::new()
+        .add_event(Sent {
+            user: ctx.contract,
+            to: ctx.sender,
+            coins: coins.clone(),
+        })?
```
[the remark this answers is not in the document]
larry0x: @Rhaki

> i think we should not emit a Sent event in recover_transfer because we are not decreasing the bank balance

I think we should:

- on `bank_execute`
  - decrease the bank's balance 
  - increase the bank's balance 
  - emit a `Sent` and `Receive` events from sender to bank
- decrease the bank's balance on `recover_transfer`
  - decrease the bank's balance
  - increase the balance of `ctx.sender` (can be either the sender or the recipient)
  - emit a `Sent` and `Receive` events from bank to sender

Issue #536, opened 2025-03-13
grug/dango: further changes to events & dango bank


larry0x:
- [ ] `ContractEvent::new` should take a `&U` instead of a `U`
- [ ] when handling an orphaned transfer, add the coin balances to the bank contract's. this ensure the sum of all balances always matches the total supply.
  - https://github.com/left-curve/left-curve/pull/537
- [ ] make `grug::Message::Transfer` a batch transfer by default. remove `dango_types::bank::ExecuteMsg::BatchTransfer` (it will just be a normal transfer).
  - https://github.com/left-curve/left-curve/pull/541

PR #537, opened 2025-03-13
fix(dango): fix orphaned transfer; native multi-send

larry0x:
closes: #536

- make `Message::Transfer` natively multi-send
- remove dango bank's `batch_transfer` method
- properly handle orphaned transfers in `mint` and `bank_execute`
- add test


larry0x:
NOTE: right now, we handle orphaned transfers in `mint`, but not in `batch_transfer` and `force_transfer`.

larry0x:
> NOTE: right now, we handle orphaned transfers in `mint`, but not in `batch_transfer` and `force_transfer`.

we have removed `batch_transfer`. the batch transfer functionality is natively part of `bank_execute`, which correctly handles orphaned transfers.

as for `force_transfer`, instead of allowing `ctx.sender` to appoint an arbitrary recipient, we can always transfer the funds to `ctx.sender`. the function can be renamed to `force_withdraw`.

PR #545, opened 2025-03-15
feat(dango): passive liquidity on order book for XYK curve

pacmanifold:
> [!IMPORTANT]
> Add passive liquidity reflection for XYK curve to Dango DEX, enabling order book integration and merging with real orders.
> 
>   - **Behavior**:
>     - Adds passive liquidity reflection onto the order book for XYK curve in `liquidity_pool.rs`.
>     - Introduces `reflect_curve()` in `liquidity_pool.rs` to generate bid and ask orders from pool reserves.
>     - Implements `MergedOrders` in `merged_orders.rs` to merge passive and real orders.
>   - **Models**:
>     - Updates `CurveInvariant::Xyk` to include `order_spacing` in `types.rs`.
>     - Modifies `PairParams` to support new curve invariant structure.
>   - **Tests**:
>     - Adds test cases in `dex.rs` to validate passive liquidity behavior and order merging.
>     - Tests cover scenarios with different fee rates and order sizes.
>   - **Misc**:
>     - Updates `clear_orders_of_pair()` in `execute.rs` to handle merged orders.
>     - Adjusts `NEXT_ORDER_ID` initialization in `state.rs` to start from 1.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=left-curve%2Fleft-curve&utm_source=github&utm_medium=referral)<sup> for ae1b1c7de8e54cb302e5912fc88b4c9e848aaab4. You can [customize](https://app.ellipsis.dev/left-curve/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


pacmanifold:
@larry0x merged main again. now uses `MergedOrders` struct, made it specific for this for now, don't really know the other use cases well enough to know how the proper abstraction should be.

The easiest way to deal with order IDs I found was to start order id counting from 1 for normal orders and let the passive liquidity add orders that all have id 0 or u64::MAX. That also has the side effect that the passive liquidity always takes priority over normal orders if put at same price. Don't know if this is the best approach but was the one requiring the least amount of changes to the existing order logic

larry0x:
1. in account querier, add to cache even if account is `None`
2. use `query_wasm_path` in account querier
3. put `order_depth` and `order_spacing` parameters into `CurveInvariant::Xyk` instead
4. `LiquidityPool::reflect_curve` method should return two iterators, instead of two `Vec`s

pacmanifold:
@larry0x addressed your above comments

dango/dex/src/core/liquidity_pool.rs
```diff
@@ -93,6 +102,49 @@ pub trait PassiveLiquidityPool {
...
         output: Coin,
     ) -> anyhow::Result<(CoinPair, Coin)>;
+
+    /// Reflect the curve onto the orderbook.
+    ///
+    /// ## Inputs
+    ///
+    /// - `base_denom`: The base asset of the pool.
+    /// - `quote_denom`: The quote asset of the pool.
+    /// - `reserves`: The current pool reserves.
```
larry0x, line 112: @pacmanifold Our terminology is that a `CoinPair` for a single pool is a "reserve" (singular). An iterator over `CoinPair`s of many pools are "reserves" (plural)

dango/dex/src/core/liquidity_pool.rs
```diff
@@ -93,6 +102,49 @@ pub trait PassiveLiquidityPool {
...
+    ///
+    /// ## Outputs
+    ///
+    /// - The spot price of the pool.
+    fn spot_price(
+        &self,
+        base_denom: &Denom,
+        quote_denom: &Denom,
+        reserves: &CoinPair,
+    ) -> StdResult<Udec128>;
```
larry0x, line 147: I realized in academic literature, this is called "marginal price". I've been using this term throughout the docs. See: https://github.com/left-curve/left-curve/tree/main/book/dex/passive-liquidity#order-placement

dango/dex/src/core/liquidity_pool.rs
```diff
@@ -254,6 +306,98 @@ impl PassiveLiquidityPool for PairParams {
...
+        let base_reserves = reserves.amount_of(&base_denom)?;
+        let quote_reserves = reserves.amount_of(&quote_denom)?;
+        let spot_price = self.spot_price(&base_denom, &quote_denom, reserves)?;
+
+        let swap_fee_rate = self.swap_fee_rate.into_inner();
+
+        // Construct the bid order iterator.
+        let mut bid_i = 1;
+        let mut bid_prev_amount = Uint128::ZERO;
+        let mut bid_price = spot_price.checked_mul(Udec128::ONE.checked_sub(swap_fee_rate)?)?;
```
larry0x, line 328: @pacmanifold I think this should be half the swap fee rate. You have half on the buy side, half on the sell side, combined the spread equals the swap fee rate.

dango/dex/src/core/liquidity_pool.rs
```diff
@@ -254,6 +306,98 @@ impl PassiveLiquidityPool for PairParams {
...
+
+    fn reflect_curve(
+        &self,
+        base_denom: Denom,
+        quote_denom: Denom,
+        reserves: &CoinPair,
+    ) -> StdResult<(
+        impl Iterator<Item = StdResult<(Udec128, Uint128)>>,
+        impl Iterator<Item = StdResult<(Udec128, Uint128)>>,
+    )> {
```
larry0x, line 318: @pacmanifold I'm thinking instead of returning `StdResult<(Udec128, Uint128)>` we just return `(Udec128, Uint128)`.

Firstly, this makes your code simpler: no need for the `unwrap_or_return_err_as_option!` macro.

Secondly, if there's an error in reflecting the curve, it only means the iterator returns `None`, meaning no more order is placed, instead of throwing an error in `cron_execute` which reverts all the changes. Generally, we should ensure `cron_execute` never fails.

Thoughts?

dango/dex/src/core/liquidity_pool.rs
```diff
@@ -263,3 +407,192 @@ fn abs_diff(a: Uint128, b: Uint128) -> Uint128 {
...
+            .reflect_curve(eth::DENOM.clone(), usdc::DENOM.clone(), &reserves)
+            .unwrap();
+
+        for (bid, expected_bid) in bids.zip(expected_bids.iter()) {
+            let bid = bid.unwrap();
+            assert_eq!(bid.0, expected_bid.0);
+            assert!(
+                bid.1.into_inner().abs_diff(expected_bid.1.into_inner()) <= order_size_tolerance
+            );
+        }
```
larry0x, line 588: Generally I don't like to do `bid.0`, `bid.1`. It's hard to tell what 0 or 1 means.

It's better if you do:

```rust
let (bid_price, bid_amount) = bid;
```

And then use `bid_price`, `bid_amount` instead of `bid.0`, `bid.1`.

dango/dex/src/execute.rs
```diff
@@ -546,13 +606,53 @@ fn clear_orders_of_pair(
...
+
+            // Convert the passive bids and asks iterators to the format required by match_orders.
+            let passive_bids = passive_bids.map(|res| {
+                res.map(|(price, amount)| {
+                    ((price, u64::MAX), Order {
+                        user: dex_addr,
+                        amount,
+                        remaining: amount,
+                        created_at_block_height: current_block_height,
+                    })
```
larry0x, line 633: If you use `current_block_height` here then the passive pool is charged the taker fee rate. I think we should charge it the maker rate instead. To do that, just use `0` for the block height.

larry0x: Nvm, this is not correct--you actually charge zero fee for the passive pool in execute.rs

pacmanifold:
- what swap fee should we charge to make instant swap and market order equivalent
- remove order_depth paramerer

pacmanifold (commented):
changed in 711c1bd817a101ed7249289dc5c578dda5370a82 lgtm

PR #546, opened 2025-03-17
feat(dex): implement volume tracking

### LINKED ISSUE #513: dango: track user's cumulative trading volume and fees paid onchain
We need to track each user's volume for the purpose such as a points campaign. We can do this offchain using indexer data, but it's better to do it onchain, as some use cases require it: e.g., if we want to base a user's trading fee on their trading volume in the last 14 days ([a common setup in CEXs](https://hyperliquid.gitbook.io/hyperliquid-docs/trading/fees)) then we need that volume data onchain.

We can create two new maps in dango-dex contract:

```rust
const VOLUMES: Map<(&Addr, Timestamp), Uint128>;

const VOLUMES_BY_USER: Map<(&Username, Timestamp), Uint128>;
```

`VOLUMES` records the cumulative volume of a user at a given timestamp, in USD. Whenever a trade is fulfilled, we fetch oracle price and update this value.

To find an account's volume in the last 14 days, we can simply do:

```rust
let current_time = ctx.block.timestamp;
let two_weeks_ago = current_time - Duration::from_weeks(2);

let volume_now = VOLUMES
    .prefix(user_addr)
    .values(ctx.storage, None, Some(current_time), Order::Descending)
    .next()
    .unwrap_or(Uint128::ZERO);
let volume_two_weeks_ago = VOLUMES
    .prefix(user_addr)
    .values(ctx.storage, None, Some(two_weeks_ago), Order::Descending)
    .next()
    .unwrap_or(Uint128::ZERO);

let volume_within_last_two_weeks = volume_now - volume_two_weeks_ago;
```

`VOLUMES_BY_USER` is the same, but aggregates the volume of all of a user's accounts. A question is how do we handle multisigs, since a multisig is associated with multiple usernames. I'm not sure.

We can also track the total amount of fee paid. This include both gas fees and trading fees and various other fees. Taxman contract may be a good place for this. We can also separate fees into different categories:

```rust
enum FeeType {
    Gas,
    Swap,
    Maker,
    Taker,
    Liquidation,
    // ...
}

const FEES: Map<(Addr, Timestamp, FeeType), Uint128>;
```

apollo-sturdy:
Partially closes #513 (fees not yet implemented)

Note: Does not cache oracle price queries. We should probably implement this in the oracle querier, since it's needed also in other places?


larry0x:
> fees not yet implemented

@apollo-sturdy there are only two ways fee can be paid:

1. in taxman's `finalize_fee` method
2. in taxman's `ExecuteMsg::Pay` method

You can just track the fees in those two methods. It doesn't matter whether a specific type of fee is already implemented or not.

apollo-sturdy:
> > fees not yet implemented
> 
> @apollo-sturdy there are only two ways fee can be paid:
> 
> 1. in taxman's `finalize_fee` method
> 2. in taxman's `ExecuteMsg::Pay` method
> 
> You can just track the fees in those two methods. It doesn't matter whether a specific type of fee is already implemented or not.

This would mean we store tracked trading fees in taxman instead of dex, and would need to add a QueryMsg to taxman. Is this desirable, or better to keep all of this in dex?

dango/dex/src/execute.rs
```diff
@@ -517,6 +523,10 @@ fn clear_orders_of_pair(
         volume,
     })?;
 
+    let dango_config = querier.query_dango_config()?;
+    let oracle = dango_config.addresses.oracle;
+    let account_factory = dango_config.addresses.account_factory;
```
larry0x, line 528: You're querying these for every trading pair. Better to query once and pass the `dango_config` as a input to the `clear_orders_of_pair` function.

dango/dex/src/execute.rs
```diff
@@ -573,6 +583,51 @@ fn clear_orders_of_pair(
                 &order,
             )?;
         }
+
+        // Calculate the volume in USD for the filled order
+        let base_asset_price = querier.query_price(oracle, &base_denom, None)?;
+        let volume = base_asset_price.value_of_unit_amount(filled)?.into_int(); // TODO: Better to store as Decimal?
```
larry0x, line 589: Yeah decimal is better.

dango/dex/src/execute.rs
```diff
@@ -573,6 +583,51 @@ fn clear_orders_of_pair(
...
+            .next()
+            .transpose()?
+            .unwrap_or(Uint128::ZERO);
+
+        // Record trading volume for the user's address
+        VOLUMES.save(
+            storage,
+            (&order.user, current_time),
+            &previous_volume.checked_add(volume)?,
+        )?;
```
larry0x, line 604: 1. If a user has multiple orders in the same block, this will write to the storage multiple times. Better if we collect the updated volumes in a `HashMap` and write once.

Btw, do you have tests for this scenario (multiple orders from the same user in multiple pairs, in the same block)?

2. We should purge volumes data that are too old. We can add a `max_volume_age` parameter (e.g. 3 weeks) to the DEX contract's config, and use `Map::clear` to remove volume data of which `timestamp + max_volume_age` is smaller than `ctx.block.timestamp`.

dango/dex/src/execute.rs
```diff
@@ -573,6 +583,51 @@ fn clear_orders_of_pair(
...
+                .transpose()?
+                .unwrap_or(Uint128::ZERO);
+
+            // Record trading volume for the user's username
+            VOLUMES_BY_USER.save(
+                storage,
+                (&username, current_time),
+                &previous_volume.checked_add(volume)?,
+            )?;
+        }
```
larry0x, line 630: Can you extract these to two separate functions? So that it looks like this:

```rust
fn clear_orders_of_pair(/* ... */) -> anyhow::Result<()> {
    // ...

    record_volume(/* ... */)?;
    record_volume_by_username(/* ... *.)
}
```

larry0x:
@apollo-sturdy I think the price query and username query both can use a querier struct with caching.

Let's do these in separate PRs.

dango/dex/src/execute.rs
larry0x: @apollo-sturdy For (1), see: 9af8bb8471d6f2ee8e3c8df262d5152490f4d924

For (2), let's do this in a separate PR.

dango/dex/src/execute.rs
larry0x: See: https://github.com/left-curve/left-curve/pull/546/commits/22186862ef7fee317d7555a6692a6f07ae37166a

dango/dex/src/execute.rs
apollo-sturdy: No I don't have any tests for that. Can add it in a separate PR!

larry0x: Yeah please

apollo-sturdy: Added here: https://github.com/left-curve/left-curve/pull/649

PR #547, opened 2025-03-17
Add a BlockTransactions dataloader

penso:
Add a dataloader to fetch all related transactions for given blocks


Cargo.toml
```diff
@@ -65,6 +65,7 @@ actix-web               = "4"
 aes-gcm                 = "0.10"
 alloy                   = "0.11"
 anyhow                  = "1"
+assert-json-diff        = "0"
```
larry0x, line 68: @penso it looks like the latest version of this package is `v2.0.2`. any reason to choose v0?

penso: none, let me update it to 2

larry0x: @penso i just did a commit to update all dependencies to latest (https://github.com/left-curve/left-curve/commit/6e52d92cd106c695df64a219d0166a6276a202c6) including bumping this one to 2, so no need for you to do anything

PR #583, opened 2025-03-26
refactor(dango): require user signature when onboarding

larry0x:
- Replace the `secret` in `NewUserSalt` to the `username`.
- In `account_factory::ExecuteMsg::RegisterUser`, user must provide the `NewUserSalt` and a signature signed over the salt.
- Change the `UserRegistered` event (@Rhaki this impacts the Galxe bot)

### DIFF
--- a/dango/account/factory/src/execute.rs
+++ b/dango/account/factory/src/execute.rs
@@ -1,6 +1,7 @@
 use {
     crate::{ACCOUNTS, ACCOUNTS_BY_USER, CODE_HASHES, KEYS, MINIMUM_DEPOSIT, NEXT_ACCOUNT_INDEX},
     anyhow::{bail, ensure},
+    dango_auth::{VerifyData, verify_signature},
     dango_types::{
         account::{self, single},
         account_factory::{
@@ -11,7 +12,7 @@ use {
     },
     grug::{
         Addr, AuthCtx, AuthMode, AuthResponse, Coins, Hash256, Inner, JsonDeExt, Message,
-        MsgExecute, MutableCtx, Op, Order, Response, StdResult, Storage, Tx,
+        MsgExecute, MutableCtx, Op, Order, Response, StdResult, Storage, Tx, json,
     },
 };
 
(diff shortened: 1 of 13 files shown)


dango/account/factory/src/execute.rs
```diff
@@ -199,24 +224,25 @@ fn onboard_new_user(
...
+        UserRegistered {
+            username,
+            key,
+            key_hash,
+        },
+        AccountRegistered {
+            address,
+            params: account.params,
+        },
+    ))
```
larry0x, line 245: @Rhaki Previously you only emit a single `UserRegistered` event here. However, we're actually doing two things here: creating the user and creating the user's first, so there should be an event for each of these two actions. Now this function returns two events.

larry0x: @Rhaki Hmm but this impacts your bot, right? Cause now you need to mark the "open a subaccount" quest as complete only **the second time** you observe an `AccountRegistered` event for a given username.

dango/testing/tests/factory.rs
```diff
@@ -334,85 +352,51 @@ fn update_key() {
...
+                    data: NewUserSalt {
+                        username: user.username.clone(),
+                        key: user.first_key(),
+                        key_hash: user.first_key_hash(),
+                    },
+                    signature: attacker
+                        .sign_arbitrary(json!({
+                            "username": "random_username",
+                        }))
+                        .unwrap(),
```
larry0x, line 395: @j0nl1 I think the more reasonable test case is: the attacker changes the username in the message, but keeps the user's signature.

larry0x: Actually let me add tests for all possible cases.

dango/auth/src/lib.rs
```diff
@@ -5,22 +5,34 @@ use {
...
+pub mod account_factory {
+    use {
+        dango_types::{account_factory::Username, auth::Key},
+        grug::{Addr, Hash256, Map, Set},
+    };
+
+    pub const KEYS: Map<(&Username, Hash256), Key> = Map::new("key");
+
+    pub const ACCOUNTS_BY_USER: Set<(&Username, Addr)> = Set::new("account__user");
+}
```
larry0x, line 34: For record: this is necessary because now, dango-account-factory contract needs to depend on dango-auth (in order to use the `verify_signature` function). Therefore, dango-auth can no longer import these storage constants from dango-account-factory, otherwise there would be a circular dependency. As such, we must define these here in dango-auth.

PR #585, opened 2025-03-27
WIP: Solidly curve implementation for passive liquidity

pacmanifold:
## TODO
- [ ] More test cases
- [ ] Reduce LUT size
- [x] Store LUT in better format
- [x] Add support for configurable spread
- [ ] Implement scaling by oracle price


pacmanifold:
> Reduce LUT size

Regarding this I ran some tests locally and increasing the LUT size to ~6000 values reduces the error to the 18th decimal place. with ~1000 values it is ~15th decimal error. 6000 values would equate to 6000*256 bit = 192 KB storage which. Increasing LUT size beyond this point does not cause any improvements, likely due to the fact that the Dec256 has 18 decimal precision

pacmanifold:
>  Add support for configurable spread

Spread and order placement now reflects the swap fee charged on instant swaps

larry0x:
Closing this because we decided to do a different type of concentrated pool.

PR #592, opened 2025-03-29
refactor(dango/auth): enforce first nonce to be close to zero but not necessarily zero

larry0x:


dango/auth/src/lib.rs
```diff
@@ -158,33 +158,35 @@ pub fn verify_nonce_and_signature(
...
+                        "nonce is too old: {} < {}",
+                        metadata.nonce,
+                        first
+                    );
+
+                    // Remove the oldest nonce if max capacity is reached.
+                    if nonces.len() == MAX_SEEN_NONCES {
+                        nonces.pop_first();
+                    }
+                } else {
```
larry0x, line 182: Changing from `match` to `if let` because I think the code looks a little nicer this way.

dango/auth/src/lib.rs
```diff
@@ -158,33 +158,35 @@ pub fn verify_nonce_and_signature(
...
+                    // Ensure the first nonce is close to zero.
+                    ensure!(
+                        metadata.nonce < MAX_NONCE_INCREASE,
+                        "first nonce is too big: {} >= MAX_NONCE_INCREASE ({})",
+                        metadata.nonce,
+                        MAX_NONCE_INCREASE
+                    );
```
larry0x, line 189: Here's the main fix of this PR

PR #603, opened 2025-04-01
dango: allow uppercase letters in usernames

larry0x:
I would like to make it possible to use uppercase letters in usernames. However we want to prevent duplicate usernames such as `larryengineer` and `LarryEngineer`. This is the same as how twitter handles work (you can't register `LarryEngineer` if `larryengineer` already exists). In order do to this, I have to introduce a `LOWERCASE_USERNAMES` set.

This PR is state breaking, so let's not merge it before the current testnet ends (in 2 weeks).


larry0x:
we decided not to do this

PR #611, opened 2025-04-03
feat(pyth): better error handling

Zexsor33:


pyth/client/src/client.rs
```diff
@@ -132,6 +132,9 @@ impl PythClientTrait for PythClient {
                                 },
                             }
                         }
+                        else {
+                            error!("Pyth SSE connection close");
+                        }
```
larry0x, line 137: This part isn't well formatted. rustfmt doesn't format contents inside macros (`stream! { ... }` here) so make sure to format them manually.

penso (approved):

PR #616, opened 2025-04-05
fix(hyperlane): add upper limit for number of signatures in ISM metadata

larry0x:
This prevents a DoS attack where an attack submits a message with a great amount of signatures.

### DIFF
--- a/hyperlane/isms/multisig/src/query.rs
+++ b/hyperlane/isms/multisig/src/query.rs
@@ -316,8 +316,8 @@ mod tests {
 
         // ----------------------- 3. Verify signatures ------------------------
 
-        // 1 signature. Should fail.
-        {
+        // 2 or 3 signatures. Shouls succeed.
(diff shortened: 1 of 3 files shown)


hyperlane/isms/multisig/src/query.rs
```diff
@@ -166,51 +186,156 @@ mod tests {
         verify(ctx.as_immutable(), &message.encode(), &raw_metadata).should_fail();
     }
 
+    /// Test three scenarios:
+    ///
+    /// 1. Number of signatures is less than threshold.
+    /// 2. Number of signatures is greater than the size of validator set.
+    /// 3. A signature is from an unknown signer (who is not in the validator set).
     #[test]
-    fn rejecting_reuse_of_signature() {
```
larry0x, line 170: This test is deleted because it's covered by the new test, `rejecting_too_many_or_too_few_or_unknown_signatures`.

PR #619, opened 2025-04-07
feat(pyth): add reconnection flow

Zexsor33:
> [!IMPORTANT]
> Adds reconnection flow with exponential backoff to Pyth client, including tests for robust SSE connection handling.
> 
>   - **Reconnection Flow**:
>     - Adds `create_event_source()` in `client.rs` to handle SSE connections with exponential backoff.
>     - Implements reconnection logic in `stream()` method in `client.rs` to handle disconnections and data absence.
>   - **Testing**:
>     - Adds `test_client_reconnection()` in `client.rs` tests to validate reconnection logic with scenarios like valid/invalid data, connection close, and server panic.
>     - Uses random port allocation for server in tests to avoid conflicts.
>   - **Dependencies**:
>     - Adds `rand` to `Cargo.toml` for random number generation in tests.
>   - **Misc**:
>     - Fixes typo in `lib.rs` ("acceptible" to "acceptable").
>     - Removes `CannotCloneRequestError` from `error.rs`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=left-curve%2Fleft-curve&utm_source=github&utm_medium=referral)<sup> for bb91d92dba7c664b535a8431e39f0cc46c6d682f. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/pyth/client/src/client.rs
+++ b/pyth/client/src/client.rs
@@ -84,7 +84,7 @@ impl PythClientTrait for PythClient {
         let keep_running = self.keep_running.clone();
 
         let stream = stream! {
-            let mut retry_attemps = 0;
+            let mut retry_attempts = 0;
 
             loop{
                 // Create an EventSource.
@@ -111,15 +111,15 @@ impl PythClientTrait for PythClient {
                             }
 
                             // Exponential backoff before try to reconnect.
-                            let delay = if retry_attemps > 6 {
+                            let delay = if retry_attempts > 6 {
                                 MAX_DELAY
                             } else {
                                 min(
                                     MAX_DELAY,
-                                    Duration::from_secs((u32::pow(2, retry_attemps)) as u64),
+                                    Duration::from_secs((u32::pow(2, retry_attempts)) as u64),
                                 )
                             };
-                            retry_attemps += 1;
+                            retry_attempts += 1;
 
                             warn!("No new data received. Reconnecting in {} seconds", delay.as_secs());
 
@@ -141,7 +141,7 @@ impl PythClientTrait for PythClient {
                                         info!("Pyth SSE connection open");
                                     },
                                     Ok(Event::Message(message)) => {
-                                        retry_attemps = 0;
+                                        retry_attempts = 0;
 
                                         match message.data.deserialize_json::<LatestVaaResponse>() {
                                             Ok(vaas) => {
--- a/pyth/client/src/client.rs
+++ b/pyth/client/src/client.rs
@@ -84,6 +84,7 @@ impl PythClient {
                         "Failed to create EventSource. Reconnecting in {} seconds",
                         current_sleep.as_secs()
                     );
+
                     sleep(current_sleep).await;
 
                     current_sleep = fn_next_backoff_duration(current_sleep);
@@ -105,6 +106,7 @@ impl PythClient {
                 Some(event_result) => match event_result {
                     Ok(_) => {
                         info!("Pyth SSE connection open");
+
                         return es;
                     },
                     Err(err) => {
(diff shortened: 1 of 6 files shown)


pyth/client/src/client.rs
```diff
@@ -76,66 +80,99 @@ impl PythClientTrait for PythClient {
...
-                    _ = tokio::time::sleep(tokio::time::Duration::from_millis(1000)) => {
-                        if !keep_running.load(Ordering::Relaxed) {
+                // Create an `EventSource`.
+                let mut es = EventSource::new(builder.try_clone().unwrap()).unwrap();
+
+                // Set the exponential backoff for reconnect.
+                es.set_retry_policy(Box::new(ExponentialBackoff::new(
+                    Duration::from_secs(1),
+                    1.5,
+                    Some(Duration::from_secs(30)),
```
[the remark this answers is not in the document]
Zexsor33: One possible idea is to have multiple price source associated with the same denom inside the oracle contract with different priority. Each time a price is updated, we save the priority and the timestamp. On new price we update if:
- the priority of new price is greater or equals than the one saved **or**
- has passed more the x seconds from the latest update.

On PP we keep retrieving prices from different oracles and submit all to the oracle contract.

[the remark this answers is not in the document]
larry0x: In the long term I would go with Pyth Lazer as primary and Pyth Core as fallback.

I don't know if any other oracle has low latency (subsecond) feed that we need.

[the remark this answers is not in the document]
Zexsor33: Yeah, Pyth + backup sources that are used just in case Pyth chain stops for some reason

pyth/client/src/client.rs
```diff
@@ -54,6 +58,83 @@ impl PythClient {
...
+            } else {
+                min(MAX_DELAY, current_sleep * 2)
+            }
+        };
+
+        let mut current_sleep = backoff_start_duration;
+
+        // Create the `EventSource`.
+        loop {
+            let mut es = match EventSource::new(builder.try_clone().unwrap()) {
```
larry0x, line 79: @Zexsor33 is `unwrap` or here?

penso: I think no important code should use `unwrap()` it's too damaging when it happens, even when it's not supposed to ever happen.

Zexsor33: To ensure the clone doesn't fail, I'm cloning it before entering the stream and return an error if it fails. I can add a error message just to be sure

penso: I think a `panic` convey a stronger message to whoever is reading this, they know it's not a lazy `unwrap`. You could also just use a `expect("Can't clone builder, something is very wrong and I can't continue.")`

penso: Just looked at the code, I see you used `let _ = builder.try_clone().ok_or(CannotCloneRequestError)?;`. I think this crate should keep doing this instead of `unwrap()` or `expect()`, that's up to whoever is calling this  code to know what to do with it, panic shouldn't be part of this crate. It's very annoying when a crate panics without you to know it will.

The calling crate/code, in our case the crate `dango-proposal-preparer` could panic or do whatever when this error has been returned.

larry0x:
@ellipsis can you review this please

pyth/client/src/client.rs
larry0x: @penso you ok with this?

penso: As said previously, I'd rather have no `unwrap`/`expect` in this crate at all and manage this error type in the code calling this.

Zexsor33: I can use ? inside the `create_event_source()`, the only problem on doing this is that the stream will return `Result<PythVaas>` instead of PythVaas. I tried to move all code that could raise Error before start the streaming, to avoid to check for errors at each iteration of `stream.next()`

penso (approved):

PR #620, opened 2025-04-07
Move `build.rs` to a `bin/` file

penso:


grug/client/src/http/schemas/schema.graphql
```diff
@@ -378,14 +378,14 @@ type TxCommitResponse {
...
 	"""
 	The base64 encoded data
 	"""
 	data: String!
 	log: String!
 	hash: String!
-	codespace: String!
 }
 
 directive @include(if: Boolean!) on FIELD | FRAGMENT_SPREAD | INLINE_FRAGMENT
```
larry0x, line 391: This file should be deleted right? The schema is used only in the indexer httpd client, not this here. This crate is the Tendermint RPC client.

PR #627, opened 2025-04-10
feat(grug): MockClient

Rhaki:
> [!IMPORTANT]
> Introduce `MockClient` for simulating blockchain client behavior and update `TestSuite` and tests to integrate with it.
> 
>   - **MockClient**:
>     - Added `MockClient` in `grug/testing/src/client.rs` to simulate blockchain client behavior.
>     - Supports `QueryClient`, `BlockClient`, `SearchTxClient`, and `BroadcastClient` traits.
>     - Handles block creation in `Timed` and `OnBroadcast` modes.
>   - **TestSuite**:
>     - Updated `TestSuite` in `grug/testing/src/suite.rs` to integrate with `MockClient`.
>     - Modified `make_block()` to return `MakeBlockOutcome`.
>   - **Outcomes**:
>     - Added `MakeBlockOutcome` struct in `grug/testing/src/outcomes.rs` to encapsulate block creation results.
>   - **Tests**:
>     - Updated tests in `dango/testing/benches/benchmarks.rs`, `dango/testing/tests/dex.rs`, and `grug/testing/tests/account.rs` to use `MockClient` and `MakeBlockOutcome`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=left-curve%2Fleft-curve&utm_source=github&utm_medium=referral)<sup> for 60e6475e646d961d7ff6a973291d99a5e383653d. It will automatically update as commits are pushed.</sup>


Cargo.toml
```diff
@@ -122,7 +122,7 @@ rand                    = "0.8" # can't use 0.9 because RustCrypto libraries sti
 reqwest                 = { version = "0.12", features = ["stream"] }
 reqwest-eventsource     = "0.6.0"
 ripemd                  = "0.1"
-rocksdb                 = "0.23"
+rocksdb                 = { version = "0.23", default-features = false, features = ["bindgen-runtime"] }
```
larry0x, line 125: The commit I showed you doesn't work here?

grug/testing/src/client.rs
```diff
@@ -0,0 +1,313 @@
...
+                .map(|(_, block)| block.clone()),
+            None => self
+                .blocks
+                .lock()
+                .await
+                .last_key_value()
+                .map(|(_, (_, block))| block.clone()),
+        };
+
+        maybe_block.ok_or(anyhow::anyhow!("block not found: {:?}", height))
```
larry0x, line 191: the message should be "block outcome not found" to differentiate with `query_block`

grug/testing/src/client.rs
```diff
@@ -0,0 +1,313 @@
...
+{
+    type Error = anyhow::Error;
+
+    async fn search_tx(&self, hash: Hash256) -> Result<SearchTxOutcome, Self::Error> {
+        self.txs
+            .lock()
+            .await
+            .get(&hash)
+            .cloned()
+            .ok_or(anyhow::anyhow!("tx not found: {:?}", hash))
```
larry0x, line 211: `Hash256` implements `Display` so you don't need to `{:?}` here

grug/testing/src/tracing.rs
```diff
@@ -3,6 +3,8 @@ use std::sync::Once;
 // The tracing subscriber can only be set once. We ensure this by using `Once`.
 static TRACING: Once = Once::new();
 
+pub use tracing::Level;
```
larry0x, line 6: Combine this `use` command with the `use` command above.

Cargo.toml
larry0x: I'm reverting this change.

grug/testing/src/client.rs
Rhaki: Yeah copilot suggested this, I didn't notice

grug/testing/src/client.rs
```diff
@@ -0,0 +1,313 @@
...
+        &self,
+        query: Query,
+        height: Option<u64>,
+    ) -> Result<QueryResponse, Self::Error> {
+        Ok(self
+            .suite
+            .lock()
+            .await
+            .app
+            .do_query_app(query, height.unwrap_or_default(), false)?)
```
larry0x, line 119: I think `.unwrap_or(0)` is more explicit in the intention of this code.

larry0x:
@Rhaki check my changes. If all good then we can merge.

grug/testing/src/client.rs
```diff
@@ -0,0 +1,313 @@
...
+) -> anyhow::Result<()>
+where
+    DB: Db,
+    VM: Vm + Clone + 'static,
+    PP: ProposalPreparer,
+    ID: Indexer,
+    AppError: From<DB::Error> + From<VM::Error> + From<PP::Error> + From<ID::Error>,
+{
+    let outcome = suite.make_block(tx);
+    index(index_txs, index_blocks, outcome, suite.block).await
```
penso, line 299: There is no need for `index` to be `async` (see other comment), meaning `next_block` doesn't need to be `async` either.

grug/testing/src/client.rs
```diff
@@ -0,0 +1,313 @@
...
+impl<DB, VM, PP, ID> SearchTxClient for MockClient<DB, VM, PP, ID>
+where
+    DB: Db + Send + Sync,
+    VM: Vm + Send + Sync,
+    PP: ProposalPreparer + Send + Sync,
+    ID: Indexer + Send + Sync,
+{
+    type Error = anyhow::Error;
+
+    async fn search_tx(&self, hash: Hash256) -> Result<SearchTxOutcome, Self::Error> {
```
penso, line 206: I don't see the need for `search_tx` to be `async` since it's only used for an asynchronous lock (**update**: ok I see it's coming from the trait definition).

Something about async locks, as explained in https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html:

> Contrary to popular belief, it is ok and often preferred to use the ordinary [Mutex](https://doc.rust-lang.org/nightly/std/sync/poison/mutex/struct.Mutex.html) from the standard library in asynchronous code.
> (...)
>  This makes the async mutex more expensive than the blocking mutex, so the blocking mutex should be preferred in the cases where it can be used.
> (...)
> Note that, although the compiler will not prevent the std Mutex from holding its guard across .await points in situations where the task is not movable between threads, this virtually never leads to correct concurrent code in practice as it can easily lead to deadlocks.

grug/testing/src/client.rs
```diff
@@ -0,0 +1,313 @@
...
+
+                let th_buffer = buffer.clone();
+                let th_suite = suite.clone();
+                let th_blocks = blocks.clone();
+                let th_txs = txs.clone();
+
+                thread::spawn(move || {
+                    let rt = Runtime::new().unwrap();
+
+                    rt.block_on(async move {
```
penso, line 54: Since this `async` is only for `lock()`, same as other comments. I think it could be removed and simplified, and `next_block` doesn't need to be `async` either. That'd remove the need for the Tokio runtime use here.

grug/testing/src/client.rs
```diff
@@ -0,0 +1,313 @@
...
+        Self {
+            suite,
+            blocks,
+            txs,
+            block_mode,
+        }
+    }
+
+    pub async fn set_timestamp(&self, timestamp: Timestamp) {
+        self.suite.lock().await.block.timestamp = timestamp;
```
penso, line 90: This `async` is only for the `lock`, same feeling about this as other. I think it can be removed and simplified.

grug/testing/src/client.rs
```diff
@@ -0,0 +1,313 @@
...
+    }
+
+    pub async fn set_block_time(&self, block_time: grug_types::Duration) {
```
penso, line 93: This `async` is only for the `lock`, same feeling about this as other. I think it can be removed and simplified.

grug/testing/src/client.rs
```diff
@@ -0,0 +1,313 @@
...
+        self.suite.lock().await.block_time = block_time;
+    }
+}
+
+#[async_trait]
+impl<DB, VM, PP, ID> QueryClient for MockClient<DB, VM, PP, ID>
```
penso, line 99: When looking at the method bodies for those, I'm wondering about the use of having two traits: `QueryClientAsync` and `QueryClient` and get `query_app_async` and `query_app` (and same for other methods). I do understand the real implementation needs `async` tho, and having 2 different traits is probably not doable as for the code calling them.

Just throwing ideas here.

grug/testing/src/client.rs
```diff
@@ -0,0 +1,313 @@
...
+                )
+                .await?;
+            },
+        }
+
+        Ok(BroadcastTxOutcome { tx_hash, check_tx })
+    }
+}
+
+async fn index(
```
penso, line 254: There is no `await` in this method body, you can remove `async`.

penso (approved):

PR #636, opened 2025-04-15
refactor(dango): various changes to lending contract

larry0x:
An "omnibus" PR for various things I want to see changed in the lending contract.

You can go through the commits one by one. The commit titles should be self explanatory.

What's missing:

- Move market and interest rate logics into `dango_lending::core`
- Use `Uint128` for `Market::total_borrowed_scaled`? (To be consistent with `total_supplied_scaled`.)
- About protocol fee--discuss whether there's better way to handle it.
- Use checked math in `InterestRateModel::calculate_rates`? (Need to discuss whether this is necessary.)
- Make variable naming consistent across the functions (especially in `dango_lending::core`).
- Make variable naming consistent: use the word "supply" everywhere, instead of "deposit".

----

> [!IMPORTANT]
> Refactors lending contract by removing `querier` parameters, updating interest rate calculations, and improving consistency in variable naming.
> 
>   - **Behavior**:
>     - Removes `querier` parameter from functions in `core.rs` files, simplifying function signatures.
>     - Updates interest rate calculations in `InterestRateModel` to use `Bounded<Udec128, ZeroInclusiveOneInclusive>` for utilization.
>     - Updates market state functions (`borrow`, `deposit`, `repay`, `withdraw`) to update interest rates after modifying market state.
>   - **Models**:
>     - Refactors `InterestRateModel` to a struct with fields for `base_rate`, `optimal_utilization`, `first_slope`, `second_slope`, and `reserve_factor`.
>     - Updates `Market` struct to include `borrow_rate` and `supply_rate` fields.
>   - **Tests**:
>     - Updates tests in `lending.rs` and `margin.rs` to reflect changes in interest rate calculations and market state updates.
>     - Adds assertions to ensure interest rates and market states are updated correctly.
>   - **Misc**:
>     - Removes `MarketUpdates` struct and replaces its usage with `InterestRateModel` directly.
>     - Fixes typos and improves comments for clarity.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=left-curve%2Fleft-curve&utm_source=github&utm_medium=referral)<sup> for 76e78da6e39ae83d88a1e2c2ecfbc57a10ec419c. It will automatically update as commits are pushed.</sup>


dango/types/src/lending/market.rs
```diff
@@ -65,10 +69,10 @@ impl Market {
         // This can happen if 100% of the supply is borrowed, which can then cause
         // borrowing to outgrow the supply due to interest accrual.
         if utilization_rate > Udec128::new_percent(100) {
-            return Ok(Udec128::new_percent(100));
+            return Ok(Bounded::new_unchecked(Udec128::new_percent(100)));
         }
 
-        Ok(utilization_rate)
+        Ok(Bounded::new_unchecked(utilization_rate))
```
Miyou, line 80: What's the point of using `Bounded` if you instantiate with unchecked?

dango/types/src/lending/market.rs
```diff
@@ -65,49 +74,40 @@ impl Market {
...
```
apollo-sturdy, line 80: What's the point of using Bounded if you instantiate with unchecked?

apollo-sturdy (commented):
lgtm.

Personally I would revert adding `total_supplied_scaled` to `Market` to stick with principle of single source of truth. Would also revert storing interest rates in Market as the computation is not heavy and it could introduce bugs if we forget to update interest rates.

dango/types/src/lending/market.rs
larry0x: If a function returns a `Bounded`, if conveys to the caller the idea that the function guarantees the result is bounded.

Although in the actual implementation of the function, we already check the value of `utilization_rate` against `0` and `1`, no need to check again, so I do `Bounded::new_unchecked`.

larry0x:
> I would revert adding `total_supplied_scaled` to `Market` to stick with principle of single source of truth. Would also revert storing interest rates in Market as the computation is not heavy and it could introduce bugs if we forget to update interest rates.

Will revert these two things

PR #648, opened 2025-04-19
feat(lending): update indices in QueryMsg::Market and QueryMsg::Markets

apollo-sturdy:
> [!IMPORTANT]
> Update `query_market` and `query_markets` to refresh indices before returning data, and adjust related test for correct flow.
> 
>   - **Behavior**:
>     - Update `query_market` and `query_markets` in `query.rs` to call `update_indices` before returning market data.
>     - Ensure indices are updated with the latest timestamp and querier data.
>   - **Testing**:
>     - Adjust test `indexes_are_updated_when_interest_rate_model_is_updated` in `lending.rs` to move time increase operation after querying market data.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=left-curve%2Fleft-curve&utm_source=github&utm_medium=referral)<sup> for 6948ac8de0e173deb00d1331737d493d2ea891c7. You can [customize](https://app.ellipsis.dev/left-curve/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/dango/lending/src/query.rs
+++ b/dango/lending/src/query.rs
@@ -1,10 +1,7 @@
 use {
     crate::{DEBTS, MARKETS, core},
     dango_types::lending::{Market, QueryMsg},
-    grug::{
-        Addr, Bound, Coins, DEFAULT_PAGE_LIMIT, Denom, ImmutableCtx, Json, JsonSerExt, Order,
-        StdResult,
-    },
+    grug::{Addr, Bound, Coins, DEFAULT_PAGE_LIMIT, Denom, ImmutableCtx, Json, JsonSerExt, Order},
     std::collections::BTreeMap,
 };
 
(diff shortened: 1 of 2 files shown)


dango/lending/src/query.rs
```diff
@@ -41,21 +38,30 @@ pub fn query(ctx: ImmutableCtx, msg: QueryMsg) -> anyhow::Result<Json> {
     .map_err(Into::into)
 }
 
-fn query_market(ctx: ImmutableCtx, denom: Denom) -> StdResult<Market> {
-    MARKETS.load(ctx.storage, &denom)
+fn query_market(ctx: ImmutableCtx, denom: Denom) -> anyhow::Result<Market> {
+    MARKETS
+        .load(ctx.storage, &denom)?
+        .update_indices(&ctx.querier, ctx.block.timestamp)
```
larry0x, line 44: @apollo-sturdy I did this PR: https://github.com/left-curve/left-curve/pull/650, now `QuerierWrapper` is `Copy`-able, so functions should use `ctx.querier` instead of `&ctx.querier`.

PR #658, opened 2025-04-22
dango(refactor): lending revamp

larry0x:
1. Use `Uint128` instead of `Udec256` for scaled debt amount.

2. Do not tokenize LP tokens; store scaled deposit amounts in the contract instead, same as debts.

   Reason 1 is related to the `ExecuteMsg::Withdraw` API. Currently we withdraw by sending LP tokens to the contract. This means we can't withdraw an exact amount of underlying tokens. Let's say I want to withdraw 100 USDC. The frontend computes 100 USDC equals 99.9 LP tokens. When the tx confirms, 99.9 LP tokens may no longer equal 100 USDC, so I get a different amount. This is bad UX.

   Reason 2 is, by putting scaled deposit amounts in the contract, we can have highly symmetrical logics between deposit<>borrow, and withdraw<>repay, which is good.

3. In deposit/withdraw/borrow/repay, we first convert the scaled amount to underlying amount, perform the addition/subtraction on the underlying amount, then convert back to scaled amount.

   This allows us to use the same rounding logic when converting between scaled and underlying amounts. For deposits we always round down; for debts we always round up. I introduced a `core::scaling` submodule that handles this.

Apparently, with these changes, the `compute_health` logic in margin account needs to be changed, as well as the LP token price source on the oracle contract. Tests need to be updated as well.

### DIFF
--- a/dango/lending/src/core/withdraw.rs
+++ b/dango/lending/src/core/withdraw.rs
@@ -1,38 +1,72 @@
 use {
-    crate::{MARKETS, core},
+    crate::{ASSETS, MARKETS, core},
     anyhow::bail,
-    dango_types::lending::{Market, NAMESPACE, SUBNAMESPACE},
-    grug::{Coin, Coins, Denom, QuerierWrapper, Storage, Timestamp},
-    std::collections::BTreeMap,
+    dango_types::lending::Market,
+    grug::{Addr, Coins, Denom, Number, Storage, Timestamp, Uint128},
+    std::{cmp::Ordering, collections::BTreeMap},
 };
 
-/// Calculates the amount of underlying coins to withdraw for a given amount of LP tokens.
-/// Returns the amount of underlying coins and the updated markets.
 pub fn withdraw(
     storage: &dyn Storage,
-    querier: QuerierWrapper,
     current_time: Timestamp,
-    coins: Coins,
-) -> anyhow::Result<(Coins, BTreeMap<Denom, Market>)> {
-    let mut withdrawn = Coins::new();
-    let mut markets = BTreeMap::new();
+    sender: Addr,
+    coins: &Coins,
+) -> anyhow::Result<(BTreeMap<Denom, Uint128>, Vec<(Denom, Market)>)> {
+    let mut markets = Vec::with_capacity(coins.len());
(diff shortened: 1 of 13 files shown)


Rhaki:
Identify with the word `assets` the collaterals makes some confusion. Isn't `collaterals` a better name?
Because functions like `scaled_asset_to_underlying` and `underlying_asset_to_scaled` need to be used only for the collateral and not for the debts. Maybe `scaled_collateral_to_underlying` is more clear

larry0x:
> Isn't `collaterals` a better name?

Not sure about this, because you're not using your deposit as collateral to borrow against.

dango/lending/src/core/withdraw.rs
```diff
@@ -1,38 +1,72 @@
...
+                Ordering::Greater => {
+                    // Decrease the user's underlying asset amount.
+                    let underlying_after = underlying_before.checked_sub(*coin.amount)?;
+
+                    // Convert the user's updated asset from underlying to scaled.
+                    let scaled_after = core::underlying_asset_to_scaled(underlying_after, &market)?;
+
+                    scaled_assets.insert(coin.denom.clone(), scaled_after);
+
+                    scaled_after - scaled_before
```
Rhaki, line 58: This should be `scaled_before - scaled_after`

larry0x: Fixed

dango/types/src/lending/market.rs
```diff
@@ -32,39 +30,18 @@ pub struct Market {
...
-    }
-
-    /// Immutably deducts the given amount from the scaled total borrowed and
-    /// returns the new market state.
-    pub fn deduct_borrowed(self, amount_scaled: Udec256) -> MathResult<Self> {
-        Ok(Self {
-            total_borrowed_scaled: self.total_borrowed_scaled.checked_sub(amount_scaled)?,
-            ..self
-        })
-    }
```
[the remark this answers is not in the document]
larry0x: I just think this syntax:

```rust
market.total_borrowed_scaled.checked_add(diff)?;
```

look better than this:

```rust
market = market.add_borrowed(diff)?;
```

Purely aesthetic reason.

Didn't remove the others because this PR doesn't touch those parts of the code.

PR #679, opened 2025-05-06
feat(grug): ClientWrapper

Rhaki:


grug/types/src/client.rs
```diff
@@ -5,3 +5,86 @@ mod query;
...
+}
+
+#[async_trait]
+impl<E, P> BroadcastClient for ClientWrapper<E, P> {
+    type Error = E;
+
+    async fn broadcast_tx(&self, tx: Tx) -> Result<BroadcastTxOutcome, Self::Error> {
+        self.client.broadcast_tx(tx).await
+    }
+}
```
larry0x, line 82: @Rhaki You've only implemented `QueryClient` and `BroadcastClient` for `ClientWrapper`. Should you also implement the other two client traits?

grug/types/src/client.rs
```diff
@@ -5,3 +5,86 @@ mod query;
...
+
+impl<E, P> Deref for ClientWrapper<E, P> {
+    type Target = dyn Client<E, P>;
+
+    fn deref(&self) -> &Self::Target {
+        self.client.as_ref()
+    }
+}
```
larry0x, line 90: Is this necessary? I mean does anything break if you remove these

grug/types/src/lib.rs
```diff
@@ -22,6 +22,7 @@ mod ffi;
 mod hash;
 mod hashers;
 mod imports;
+mod jellifish_merkle;
```
larry0x, line 25: @Rhaki Typo, should be `jellyfish`

PR #692, opened 2025-05-09
feat(dango): Account Factory query users by key hash

Rhaki:
> [!IMPORTANT]
> Add `ForgotUsername` query to retrieve usernames by key hash, updating relevant functions and state management.
> 
>   - **Behavior**:
>     - Adds `ForgotUsername` query to `QueryMsg` in `msg.rs` to retrieve usernames by key hash.
>     - Implements `forgot_username()` in `query.rs` to handle the new query.
>     - Updates `instantiate()`, `register_user()`, and `update_key()` in `execute.rs` to manage `USERNAMES_BY_KEY`.
>   - **State**:
>     - Introduces `USERNAMES_BY_KEY` in `state.rs` to map key hashes to usernames.
>   - **Misc**:
>     - Updates imports in `execute.rs`, `query.rs`, and `msg.rs` to include `USERNAMES_BY_KEY` and `BTreeSet`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=left-curve%2Fleft-curve&utm_source=github&utm_medium=referral)<sup> for f03722e15fc53a78193581df361c728b97b6537d. You can [customize](https://app.ellipsis.dev/left-curve/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/dango/types/src/account_factory/msg.rs
+++ b/dango/types/src/account_factory/msg.rs
@@ -113,8 +113,10 @@ pub enum QueryMsg {
     /// Query a single user by username.
     #[returns(User)]
     User { username: Username },
+    /// Query usernames associated with a given key hash.
+    /// Useful if user forgot their username but still have access to the key.
     #[returns(BTreeSet<Username>)]
-    UsersByKey {
+    ForgotUsername {
         key_hash: Hash256,
         start_after: Option<Username>,
         limit: Option<u32>,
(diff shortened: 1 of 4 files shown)


dango/types/src/account_factory/msg.rs
```diff
@@ -113,6 +113,12 @@ pub enum QueryMsg {
     /// Query a single user by username.
     #[returns(User)]
     User { username: Username },
+    #[returns(BTreeSet<Username>)]
+    UsersByKey {
```
larry0x, line 117: I'm going to rename this query. We have a query called `QueryMsg::User` that returns a `User` struct. If you call this query `UsersByKey` then I would expect it also returns `User` structs.

To avoid confusion, we can either rename this to `UsernamesByKey` or just `ForgotUsername`.

PR #713, opened 2025-05-16
refactor(dango): account querier to return `Option<&Account>`

larry0x:
A contract isn't necessarily an account registered at the account factory. So it makes sense to return a `None` in case it's not.

----

> [!IMPORTANT]
> Refactor `AccountQuerier` to return `Option<&Account>` and update related functions to handle unregistered accounts.
> 
>   - **Behavior**:
>     - `AccountQuerier` in `account_querier.rs` now returns `Option<&Account>` in `query_account()` to handle unregistered accounts.
>     - `may_query_wasm_path` in `querier.rs` now takes `&Path` instead of `Path`.
>   - **Usage**:
>     - Updated `clear_orders_of_pair()` in `execute.rs` to handle `Option<&Account>` when recording trading volume.
>     - Updated `query_health()` in `core.rs` to use `&Path` in `may_query_wasm_path()`.
>   - **Misc**:
>     - Minor refactor in `account_querier.rs` to convert `&Option<T>` to `Option<&T>`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=left-curve%2Fleft-curve&utm_source=github&utm_medium=referral)<sup> for e941dea93a5364bee62d44d739d7868dd232241a. You can [customize](https://app.ellipsis.dev/left-curve/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/dango/account/factory/src/account_querier.rs
+++ b/dango/account/factory/src/account_querier.rs
@@ -5,19 +5,19 @@ use {
 };
 
 pub struct AccountQuerier<'a> {
-    cache: Cache<'a, Addr, Account>,
+    cache: Cache<'a, Addr, Option<Account>>,
 }
 
 impl<'a> AccountQuerier<'a> {
     pub fn new(factory: Addr, querier: QuerierWrapper<'a>) -> Self {
         Self {
             cache: Cache::new(move |address, _| {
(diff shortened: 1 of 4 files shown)


larry0x:
cc @pacmanifold, you did this in the DEX PR, but I feel it's better to do it here separately.

PR #726, opened 2025-05-19
Enable sentry and add tags

penso:


dango/cli/src/main.rs
```diff
@@ -85,7 +85,23 @@ async fn main() -> anyhow::Result<()> {
...
+            .with(fmt::layer().with_filter(max_level))
+            .with(sentry_layer())
+            .init();
+
         tracing::info!("Sentry initialized");
+    } else {
+        // Set up tracing.
+        tracing_subscriber::fmt()
+            .with_max_level(cfg.log_level.parse::<LevelFilter>()?)
+            .init();
```
larry0x, line 104: We don't set up tracing if sentry is enabled?

larry0x (approved):

PR #777, opened 2025-05-29
tracing: suppress tracing messages from actix-web and async-graphql

larry0x:


larry0x:
chatgpt made this PR which turns out to be completely incorrect. replaced by #779

PR #796, opened 2025-06-04
feat(grug): Malachite

Rhaki:
> [!IMPORTANT]
> Introduce the Malachite feature to the `grug` project, adding new actors, state management, configuration, types, and tests for enhanced consensus and mempool operations.
> 
>   - **Actors**:
>     - Add `Host`, `Mempool`, and `MempoolNetwork` actors in `host.rs`, `mempool.rs`, and `network.rs` respectively for handling consensus and mempool operations.
>     - Implement `Actor` trait for these actors to manage lifecycle and message handling.
>   - **State Management**:
>     - Introduce `State` struct in `state.rs` to manage consensus state, including height, round, and proposer information.
>     - Add `DECIDED_BLOCK`, `UNDECIDED_BLOCK`, and `UNDECIDED_PROPOSALS` maps for tracking block states.
>   - **Configuration**:
>     - Define `ActorsConfig`, `HostConfig`, and `MempoolConfig` in `config.rs` for node configuration.
>     - Add TOML configuration files for nodes in `tests/nodes_config/`.
>   - **Types and Traits**:
>     - Define new types such as `Address`, `Block`, `Proposal`, `Vote`, and `Value` in `types/`.
>     - Implement necessary traits for these types to integrate with the consensus and mempool systems.
>   - **Testing**:
>     - Add integration tests in `multiple.rs` and `single.rs` to test multi-node and single-node setups.
>     - Provide utility functions in `utils.rs` for setting up test environments.
>   - **Miscellaneous**:
>     - Add `ctx!` macro in `macros.rs` for context management.
>     - Update `Cargo.toml` and `Cargo.lock` with new dependencies for the Malachite feature.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=left-curve%2Fleft-curve&utm_source=github&utm_medium=referral)<sup> for 75458ac47a1336e0952ad029cb87dda596b7a58d. You can [customize](https://app.ellipsis.dev/left-curve/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


nenadmilosevic95 (commented):
Hi everyone, I did a full pass on mempool.rs — it looks good overall to me, nice work! I’ve left a few comments that might be worth a look.

One general note: I liked that you encapsulated the application logic into the MempoolApp and HostApp traits. We're currently thinking a lot about how to achieve a clean separation of the application logic. If you have more insights or thoughts on this, they’d be very helpful for us going forward.

grug/malachite/src/actors/mempool/mempool.rs
```diff
@@ -0,0 +1,266 @@
...
+};
+
+pub type MempoolMsg = Msg;
+pub type MempoolActorRef = ActorRef<Msg>;
+
+#[derive(Default)]
+pub struct State {
+    txs: VecDeque<RawTx>,
+    tx_hashes: HashMap<Hash256, usize>,
+}
```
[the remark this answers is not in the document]
Rhaki: Yeah, this is one of the best solution to to minimize the computation for removing properly

grug/malachite/src/actors/mempool/mempool.rs
```diff
@@ -0,0 +1,266 @@
...
+            // we are not removing the tx from the mempool, here because prepare proposal could not
+            // include some txs. Txs will be removed during decided.
+
+            max_tx_bytes = max_tx_bytes.saturating_sub(tx.0.len());
+
+            if max_tx_bytes == 0 {
+                break;
+            }
+
+            txs.push(tx.clone());
```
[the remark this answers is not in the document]
nenadmilosevic95: Yes, but why shouldn't we add a transaction that is of size 100 bytes if we have 100 bytes left?

grug/malachite/src/actors/mempool/mempool.rs
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    #[tracing::instrument("gossip_tx", skip_all)]
+    fn gossip_tx(&self, tx: RawTx) -> ActorResult<()> {
+        // TODO: Actually MempoolTransactionBatch is in prost format
+        let tx = tx.to_borsh_vec()?;
+        let msg = Any {
+            type_url: "type.googleapis.com/malachite.MempoolTransactionBatch".to_string(),
+            value: tx,
+        };
```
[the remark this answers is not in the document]
Rhaki: Sadly `MempoolTransactionBatch` require a `prost_types::Any`. We don't use protobuf format at all so right now this is a workaround

grug/malachite/src/actors/host/host.rs
```diff
@@ -0,0 +1,560 @@
...
+    ) -> ActorResult<()> {
+        let height = DECIDED_BLOCK
+            .keys(state, None, None, grug::Order::Descending)
+            .next()
+            .transpose()?
+            .unwrap_or_default();
+
+        state.set_consensus(consensus.clone());
+
+        // TODO: is a sleep necessary here? It's present in the example implementation
```
ancazamfir, line 208: I put it in the example as it was also in our other test apps. I think it was because of some timing issues (might be from testing framework), but I asked to get clarifications.

grug/malachite/src/actors/host/host.rs
```diff
@@ -0,0 +1,560 @@
...
+        &self,
+        state: &mut State,
+        _height: ctx!(Height),
+        _round: Round,
+        valid_round: Round,
+        _address: ctx!(Address),
+        block_hash: ctx!(Value::Id),
+    ) -> ActorResult<()> {
+        if let Some(data) = UNDECIDED_BLOCK.may_load(state, block_hash)? {
+            // TODO: Should we assert height, round and proposer?
```
ancazamfir, line 274: For height is fine to assert.
Why do you need a round in the block? If this is a re-proposal of a valid value you need to assert on `valid_round`. When restream is triggered by the liveness property (`valid_round` is not set), you would assert on `round`. 
Proposer may be different than the one you have in a block so don't assert on that.

grug/malachite/src/actors/host/host.rs
```diff
@@ -0,0 +1,560 @@
...
+
+            let pre_block = PreBlock::new(
+                height,
+                self.address,
+                round,
+                Timestamp::from_nanos(
+                    std::time::SystemTime::now()
+                        .duration_since(std::time::UNIX_EPOCH)
+                        .unwrap()
+                        .as_nanos(),
```
ancazamfir, line 338: For proposer based timestamp there are a few checks required.
Check the timestamp of the previous block and if higher than current time then sleep for the difference. We want the blocks to have increasing timestamps.
The other check is at proposal receive time, added a comment there.

grug/malachite/src/actors/host/host.rs
```diff
@@ -0,0 +1,560 @@
...
+        part: StreamMessage<ctx!(ProposalPart)>,
+        reply_to: RpcReplyPort<ProposedValue<Context>>,
+    ) -> ActorResult<()> {
+        // Since we are using a single part proposal, we only need data
+        let Some(ProposalData { block, valid_round }) = part.content.into_data() else {
+            return Ok(());
+        };
+
+        info!(height = %block.height, round = %block.round, "All parts received");
+
```
ancazamfir, line 400: Check that the received block timestamp is after the previous block timestamp. If not the proposal should be marked `Invalid` and sent to consensus. 
You can check the comet implementation [here](https://github.com/cometbft/cometbft/blob/bd1bc3bb6392a973d7b60fd0b19da6e71ca09b47/types/proposal.go#L85-L107).
It is a bit more complex as it takes into account clock drifts and message delays which you may want to add at some point.

grug/malachite/src/actors/host/host.rs
```diff
@@ -0,0 +1,560 @@
...
+        // Run FinalizeBlock
+        let block_hash = state.finalize_block(&block)?;
+
+        let value = ProposedValue {
+            height: block.height,
+            round: block.round,
+            valid_round,
+            proposer: block.proposer,
+            // TODO: Is this correct? If block hashes are different, which one we should use?
+            value: <ctx!(Value)>::new(block.block_hash()),
```
ancazamfir, line 410: It is correct, should be one from the received block and not the one you compute locally (`block_hash`)

grug/malachite/src/actors/host/host.rs
```diff
@@ -0,0 +1,560 @@
...
+            UNDECIDED_PROPOSALS.save(
+                state,
+                (*block.height, block.round.as_i64(), block.block_hash()),
+                &value,
+            )?;
+
+            // Store undecided block
+            UNDECIDED_BLOCK.save(state, block.block_hash(), &block)?;
+        } else {
+            warn!(block_hash = %block_hash, proposal_block_hash = %block.block_hash(), "Block hash mismatch");
```
ancazamfir, line 426: maybe `error!`?

grug/malachite/src/actors/host/host.rs
```diff
@@ -0,0 +1,560 @@
...
+        }
+
+        // TODO: If block_hash is different from the proposed value, what we should do?
```
ancazamfir, line 429: One reason to store is to handle restarts. The other is to better handle non-determinism and in general to have same code for both valid and invalid proposals up to commit. See comments on `decided()`

grug/malachite/src/actors/host/host.rs
```diff
@@ -0,0 +1,560 @@
...
+    #[tracing::instrument("decided", skip_all, fields(height = %certificate.height, round = %certificate.round))]
+    async fn decided(
+        &self,
+        state: &mut State,
+        certificate: CommitCertificate<Context>,
+        consensus: ConsensusRef<Context>,
+    ) -> ActorResult<()> {
+        let Some(block) = UNDECIDED_BLOCK.may_load(state, certificate.value_id)? else {
+            warn!(value_id = %certificate.value_id, "Proposed block not found");
+            return Err(anyhow::anyhow!("Proposed block not found").into());
```
ancazamfir, line 448: If you don't store invalid blocks and proposals this will return and non-determinism will not be flagged explicitly.

If you store invalid blocks and proposals things are a bit more clear.
I this case I would pull here also the undecided proposal that includes the validity. 
If valid you can do what the rest of the function does, including commit.

If invalid, it means > 2/3rds have decided on what this nodes thinks is an invalid value. This is a very clear non-determinism issue and you could:
- panic
- issue error, cleanup and `RestartHeight()`
   - immediately this will likely trigger sync which will bring the same invalid block. So the symptom will be a stuck node.
   - once the determinism bug is fixed the last height might need to be restarted anyway (if >2/3rds have committed a bad block)

grug/malachite/src/actors/host/state.rs
```diff
@@ -0,0 +1,218 @@
...
+        self.consensus
+            .clone()
+            .ok_or(anyhow::anyhow!("Consensus not set").into())
+    }
+
+    pub fn stream_id(&self) -> StreamId {
+        let mut bytes = Vec::with_capacity(size_of::<u64>() + size_of::<u32>());
+        bytes.extend_from_slice(&self.height.to_be_bytes());
+        // TODO: can this panic?
+        bytes.extend_from_slice(&self.round.as_u32().unwrap().to_be_bytes());
```
ancazamfir, line 71: Not in practice given the message flow. If not defined (`Round::Nil`) it will panic, otherwise the round is `u32`.
But maybe nice to anyhow instead of unwrap.

grug/malachite/src/actors/host/state.rs
```diff
@@ -0,0 +1,218 @@
...
+            if pending == block.block_hash() {
+                rerun = false;
+            }
+        }
+
+        if rerun {
+            let block_hash = self.finalize_block(block)?;
+            if block_hash != block.block_hash() {
+                // TODO: We should panic here? This mean the the we have some non-deterministic behavior in the app.
+                panic!("Block hash mismatch");
```
ancazamfir, line 128: Not sure about the `pending_commit_block_hash` use here as a trigger for rerunning `finalize_block`. It is possible to get different proposals for same or different rounds. These are all finalized in earlier steps and information is stored in `UNDECIDED_BLOCK` and `UNDECIDED_PROPOSALS`, with the latter including the validity which is basically the check of the block_hash. 
So the caller can look in `UNDECIDED_PROPOSALS` and call `commit` with a valid block.
Added related comments in `decided()`.

ancazamfir (commented):
Did a first pass on the host files, left some comments and questions.

larry0x:
No longer planned.

PR #817, opened 2025-06-11
Add metrics

penso:
> [!IMPORTANT]
> Integrate Prometheus and Grafana for metrics collection and monitoring across the application, with configurations and tests for HTTP and GraphQL metrics.
> 
>   - **Metrics Integration**:
>     - Add `metrics`, `metrics-exporter-prometheus`, and `prometheus` dependencies in `Cargo.toml` files for various components.
>     - Implement `MetricsExtension` in `indexer/httpd/src/graphql/extensions/metrics.rs` for GraphQL metrics.
>     - Add `HttpMetrics` middleware in `indexer/httpd/src/middlewares/metrics.rs` for HTTP metrics.
>     - Initialize metrics in `dango/cli/src/start.rs` and `dango/cli/src/indexer.rs`.
>   - **Configuration**:
>     - Update `app.toml` files in `deploy/roles/dango/templates` to configure metrics HTTP server.
>     - Add metrics-related configurations in `docker-compose.yml` and monitoring files under `networks/localdango/monitoring`.
>   - **Monitoring Setup**:
>     - Add Prometheus and Grafana services in `docker-compose.yml`.
>     - Configure Prometheus with `prometheus.yml` and alert rules in `alerts/graphql.yml`.
>     - Set up Grafana dashboards in `grafana/provisioning/dashboards`.
>   - **Testing**:
>     - Add `metrics.rs` test in `dango/testing/tests/httpd` to verify metrics availability.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=left-curve%2Fleft-curve&utm_source=github&utm_medium=referral)<sup> for 0c83617505e26055ce62efe34e00243bd6e32221. You can [customize](https://app.ellipsis.dev/left-curve/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


dango/cli/src/main.rs
```diff
@@ -93,25 +94,45 @@ async fn main() -> anyhow::Result<()> {
...
         tracing_subscriber::registry()
-            .with(tracing_subscriber::fmt::layer().with_filter(filter))
+            .with(
+                tracing_subscriber::EnvFilter::try_from_default_env()
+                    .unwrap_or_else(|_| "info".into()), // Default to info if RUST_LOG not set
+            )
+            .with(
+                tracing_subscriber::fmt::layer()
+                    // .with_span_events(tracing_subscriber::fmt::format::FmtSpan::CLOSE),
+            )
```
larry0x, line 115: @penso Here you removed the `filter`. It is a custom filter that removes tracing events from actix and async-graphql. Without it, we get a bunch of events on every httpd query, making the logs basically unreadable. Can you add `filter` back?

dango/cli/src/main.rs
```diff
@@ -93,25 +94,45 @@ async fn main() -> anyhow::Result<()> {
...
-        Command::Indexer(cmd) => cmd.run(app_dir).await,
+        Command::Indexer(cmd) => cmd.run(app_dir, metrics_handle).await,
         Command::Keys(cmd) => cmd.run(app_dir.keys_dir()),
         Command::Query(cmd) => cmd.run(app_dir).await,
-        Command::Start(cmd) => cmd.run(app_dir).await,
+        Command::Start(cmd) => cmd.run(app_dir, metrics_handle).await,
         #[cfg(feature = "testing")]
         Command::Test(cmd) => cmd.run().await,
         Command::Tx(cmd) => cmd.run(app_dir).await,
     }
```
larry0x, line 131: @penso It looks like only `Command::Indexer` and `Command::Start` need the `metrics_handle`. Do we need to init it even if `cli.command` is a different variant?

dango/cli/src/indexer.rs
```diff
@@ -31,10 +35,15 @@ enum SubCmd {
         /// End height (inclusive)
         end: u64,
     },
+    MetricsHttpd,
```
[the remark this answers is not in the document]
penso: Yes, when I want to run the local httpd to see what's in the indexer database.

dango/cli/src/main.rs
penso: Because it enforces the filtering, which is annoying because I then can't use `RUST_LOG=...` which is standard way to filter out logs so I change log level for a specific crate, and get more logs when debugging.

dango/cli/src/main.rs
penso: We could probably init in in each subcommand if you'd like that.

larry0x: Can you see if this works: https://github.com/left-curve/left-curve/pull/817/commits/0772c5e0d9a3d27b2d62ee9ed77683bf0b3692d8

dango/cli/src/main.rs
larry0x: In this case we can delete the custom filter, but you should set up `RUST_LOG` in deployment to hide the actix and async-graphql events. Is this ok?

PR #820, opened 2025-06-12
fix: genesis state

larry0x:


larry0x:
cc @penso, devnet is fixed.

PR #834, opened 2025-06-17
dex: use oracle prices to determine how many LP tokens to mint

larry0x:
> [!IMPORTANT]
> Use oracle prices to determine USD value for minting LP tokens and apply fees for asymmetrical deposits in DEX.
> 
>   - **Behavior**:
>     - Use oracle prices to determine USD value of deposits for minting LP tokens in `add_initial_liquidity()` and `add_subsequent_liquidity()` in `geometric.rs`.
>     - Apply deposit fee for asymmetrical deposits based on oracle prices in `add_liquidity()` in `liquidity_pool.rs`.
>   - **Functions**:
>     - Add `oracle_value()` in `geometric.rs` to calculate USD value of `CoinPair` using `OracleQuerier`.
>     - Modify `add_liquidity()` in `liquidity_pool.rs` to handle initial liquidity with oracle prices.
>   - **Tests**:
>     - Update tests in `dex.rs` to register oracle prices and validate new LP token minting logic.
>     - Add test cases for asymmetrical deposit fee logic and oracle price integration.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=left-curve%2Fleft-curve&utm_source=github&utm_medium=referral)<sup> for 990f0b360d1ec6e2b3f5965cdd8ac7ad97dd0f22. You can [customize](https://app.ellipsis.dev/left-curve/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/dango/dex/src/core/liquidity_pool.rs
+++ b/dango/dex/src/core/liquidity_pool.rs
@@ -7,9 +7,8 @@ use {
         Coin, CoinPair, Denom, IsZero, MultiplyFraction, Number, NumberConst, StdResult, Udec128,
         Uint128,
     },
+    std::ops::Sub,
 };
-
-const HALF: Udec128 = Udec128::new_percent(50);
 
 pub trait PassiveLiquidityPool {
     /// Provide liquidity to the pool. This function mutates the pool reserves.
@@ -140,12 +139,27 @@ impl PassiveLiquidityPool for PairParams {
         lp_token_supply: Uint128,
         deposit: CoinPair,
     ) -> anyhow::Result<(CoinPair, Uint128)> {
+        // The deposit must have the same denoms as the reserve. This should
+        // have been caught earlier in `execute::provide_liquidity`.
+        // We assert this in debug builds only.
+        #[cfg(debug_assertions)]
+        {
+            let deposit_denoms = (deposit.first().denom, deposit.second().denom);
+            let reserve_denoms = (reserve.first().denom, reserve.second().denom);
+            assert_eq!(
+                deposit_denoms, reserve_denoms,
+                "deposit denoms {deposit_denoms:?} don't match reserve denoms {reserve_denoms:?}",
+            );
+        }
+
         // If there isn't any liquidity in the pool yet, run the special logic
         // for adding initial liquidity, then early return.
         if lp_token_supply.is_zero() {
             let mint_amount = match &self.pool_type {
                 PassiveLiquidity::Xyk { .. } => xyk::add_initial_liquidity(&deposit)?,
-                PassiveLiquidity::Geometric { .. } => geometric::add_initial_liquidity(&deposit)?,
+                PassiveLiquidity::Geometric { .. } => {
+                    geometric::add_initial_liquidity(oracle_querier, &deposit)?
+                },
             };
 
             reserve.merge(deposit.clone())?;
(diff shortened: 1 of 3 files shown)


dango/dex/src/core/liquidity_pool.rs
```diff
@@ -162,27 +176,67 @@ impl PassiveLiquidityPool for PairParams {
...
+            let reserve_b = price.value_of_unit_amount(*reserve.second().amount)?;
+
+            let deposit_value = a.checked_add(b)?;
+            let reserve_value = reserve_a.checked_add(reserve_b)?;
+
+            abs_diff(a.checked_mul(reserve_b)?, b.checked_mul(reserve_a)?)
+                .checked_div(deposit_value.checked_add(reserve_value)?)?
+                .checked_mul(*self.swap_fee_rate)?
+                .checked_div(deposit_value)?
+        };
```
larry0x, line 232: In terms of code styling, I like using curly braces like this to section the code. I think the readability is better this way.

dango/dex/src/core/liquidity_pool.rs
```diff
@@ -296,7 +350,10 @@ impl PassiveLiquidityPool for PairParams {
 }
 
 /// Compute `|a - b|`.
-fn abs_diff(a: Uint128, b: Uint128) -> Uint128 {
+fn abs_diff<T>(a: T, b: T) -> <T as Sub>::Output
+where
+    T: PartialOrd + Sub, // TODO: do we use PartialOrd or Ord?
+{
```
larry0x, line 356: This allows us to use this method for both `Uint128` and `Udec128`.

I don't really know the difference between `PartialOrd` and `Ord` though.

pacmanifold: GPT o3's take below. Makes sense to me. I don't think an absdiff makes sense for a type that does not quarantee a complete ordering

## Use Ord

```rust
Copy
Edit
use core::ops::Sub;

/// Compute `|a − b|`.
fn abs_diff<T>(a: T, b: T) -> <T as Sub>::Output
where
    T: Ord + Sub<Output = T>,
{
    if a > b { a - b } else { b - a }
}
```

### Why Ord instead of PartialOrd?
What the branch if a > b { … } else { … } assumes	Which trait guarantees it
For every pair (a, b) exactly one of a > b, a == b, or a < b is true (a total order).	Ord 
doc.rust-lang.org

PartialOrd only promises that you might be able to compare two values; it admits “incomparable” pairs where all of those relations are false.
A classic example is NaN for f32/f64—every comparison with a NaN evaluates to false 
github.com
. If you let abs_diff accept a type that can behave like that:

```rust
let nan = f32::NAN;
abs_diff(nan, 1.0);        // `a > b` is false ⇒ executes `1.0 - nan` ⇒ NaN
```

The function still returns something, but the branch it took had nothing to do with which operand was “larger”; it was chosen because the comparison was unordered. That silently violates the intuitive contract “return the absolute difference”.

By requiring Ord you:

Reject types that cannot supply a total order (plain floats, sets, graphs with partial ordering, …).

Make the pre-condition explicit to callers.

Align with the standard-library precedent: every generic abs_diff the std library ships (for integers, Duration, etc.) constrains the type with Ord + Sub.

If you really do need to support partially-ordered types you can, but you must handle the “unordered” (None) case explicitly:

```rust
fn abs_diff_partial<T>(a: T, b: T) -> T::Output
where
    T: PartialOrd + Sub<Output = T>,
{
    match a.partial_cmp(&b) {
        Some(core::cmp::Ordering::Greater) => a - b,
        Some(_)                            => b - a,
        None                               => (a - a) - (b - b), // domain specific: maybe NaN, maybe panic
    }
}
```

For the usual numeric, totally-ordered types a simple Ord bound is safer, clearer, and what most Rust code (and the standard library) opts for.

pacmanifold (approved):
based on my abvove comment i will update abs_diff to use Ord instead

PR #851, opened 2025-06-21
Add 8601 datetime

penso:
> [!IMPORTANT]
> Add ISO 8601 datetime handling across entity models, GraphQL schemas, and tests for consistent datetime serialization.
> 
>   - **Datetime Handling**:
>     - Added `serde_iso8601` module for ISO 8601 datetime serialization/deserialization in `indexer/sql/src/serde_iso8601.rs`.
>     - Updated `Timestamp` in `grug/types/src/time.rs` to support conversion to/from `DateTime<Utc>` and `NaiveDateTime`.
>     - Added `Iso8601DateTime` scalar in `indexer/sql/src/scalars.rs` for GraphQL.
>   - **Entity Model Updates**:
>     - Updated `created_at` fields in `accounts.rs`, `public_keys.rs`, `transfers.rs`, `users.rs`, `blocks.rs`, `events.rs`, `messages.rs`, and `transactions.rs` to use `serde_iso8601` for datetime serialization.
>     - Added `created_at()` method in GraphQL `ComplexObject` implementations to return ISO 8601 formatted datetime.
>   - **GraphQL Schema Changes**:
>     - Updated `schema.graphql` to use `String` for `createdAt` fields with ISO 8601 format in `Block`, `Event`, `Message`, and `Transaction` types.
>   - **Testing**:
>     - Added `graphql_returns_iso_8601` test in `indexer/testing/tests/graphql/datetime.rs` to verify ISO 8601 datetime format in GraphQL responses.
>   - **Dependencies**:
>     - Added `chrono` and `test-case` dependencies in `Cargo.lock` and `Cargo.toml` files.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=left-curve%2Fleft-curve&utm_source=github&utm_medium=referral)<sup> for 075f833ac13314a81bdbcfd686c4b5e087917e6b. You can [customize](https://app.ellipsis.dev/left-curve/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/indexer/client/src/schemas/schema.graphql
+++ b/indexer/client/src/schemas/schema.graphql
@@ -65,7 +65,7 @@ type Event {
 	data: JSON!
 	blockHeight: Int!
 	"""
-	Returns the event's timestamp in ISO 8601 format with time zone.
+	Returns the event timestamp in ISO 8601 format with time zone.
 	"""
 	createdAt: String!
 	transaction: Transaction
--- a/indexer/sql/src/entity/events.rs
+++ b/indexer/sql/src/entity/events.rs
@@ -123,7 +123,7 @@ pub struct Model {
 #[cfg(feature = "async-graphql")]
 #[ComplexObject]
 impl Model {
-    /// Returns the event's timestamp in ISO 8601 format with time zone.
+    /// Returns the event timestamp in ISO 8601 format with time zone.
     async fn created_at(&self) -> String {
         Timestamp::from(self.created_at).to_rfc3339_string()
     }


indexer/httpd/src/graphql/scalars.rs
```diff
@@ -0,0 +1,52 @@
...
+    fn from(dt: chrono::NaiveDateTime) -> Self {
+        Self(dt)
+    }
+}
+
+impl From<Iso8601DateTime> for chrono::NaiveDateTime {
+    fn from(dt: Iso8601DateTime) -> Self {
+        dt.0
+    }
+}
```
larry0x, line 52: @penso This file appears to be not used at all?

indexer/sql/src/scalars.rs
```diff
@@ -0,0 +1,52 @@
...
+                    // Parse ISO format with microseconds but no timezone, assume UTC
+                    Ok(Self(dt))
+                } else {
+                    Err(async_graphql::InputValueError::custom(
+                        "Invalid datetime format",
+                    ))
+                }
+            },
+            _ => Err(async_graphql::InputValueError::custom("Expected string")),
+        }
```
larry0x, line 32: @penso This logic is duplicate as in `serde_iso8601.rs`. I'm going to reuse the function to avoid the duplication. Also I'm going to use the early return pattern to make the code look nicer.

indexer/httpd/src/graphql/scalars.rs
larry0x: This file appears to be a duplicate of `indexer/sql/src/scalars.rs`, which is actually used. I'm deleting it.

PR #845, opened 2025-06-20
[LEF-157] add staleness check in oracle querier

larry0x:
> [!IMPORTANT]
> Add staleness check to `OracleQuerier` to ensure prices are not older than `MAX_ORACLE_STALENESS`, updating relevant modules and adding tests.
> 
>   - **Behavior**:
>     - Introduces staleness check in `OracleQuerier` to ensure prices are not older than `MAX_ORACLE_STALENESS`.
>     - Updates `swap_exact_amount_in` in `router.rs` to use `OracleQuerier` with staleness check.
>     - Updates `cron_execute` in `cron.rs` to use `OracleQuerier` with staleness check.
>     - Updates `provide_liquidity` and `swap_exact_amount_in` in `execute.rs` to use `OracleQuerier` with staleness check.
>     - Updates `query_simulate_swap_exact_amount_in` in `query.rs` to use `OracleQuerier` with staleness check.
>   - **Constants**:
>     - Adds `MAX_ORACLE_STALENESS` constant in `lib.rs` to define maximum staleness duration.
>   - **Tests**:
>     - Adds tests in `oracle_querier.rs` to verify staleness check functionality in `OracleQuerier`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=left-curve%2Fleft-curve&utm_source=github&utm_medium=referral)<sup> for 58035e9cc3be485551aac87f7047b91650704402. You can [customize](https://app.ellipsis.dev/left-curve/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

PR #868, opened 2025-06-25
Passive order type

larry0x:


dango/testing/tests/dex_proptests.rs
```diff
@@ -93,8 +96,9 @@ fn register_fixed_price(
                 denom => dango_types::oracle::PriceSource::Fixed {
                     humanized_price,
                     precision,
-                    timestamp: 0,
-                }
+                    // Use a very recent time to avoid the "price is too old" error.
+                    timestamp: MOCK_GENESIS_TIMESTAMP,
+                },
```
larry0x, line 101: @apollo-sturdy We need to do this because of #845.

PR #875, opened 2025-06-29
chore: address a new lint introduced in rust 1.88

larry0x:
> [!IMPORTANT]
> Update string formatting to new syntax across multiple files for Rust 1.88 compatibility.
> 
>   - **String Formatting**:
>     - Update to use new format string syntax in `indexer.rs`, `margin.rs`, `abci.rs`, `main.rs`, `dec.rs`, `int.rs`, `prefix.rs`, `reply.rs`, `coin_pair.rs`, `environment.rs`, `client.rs`, `build_graphql_schema.rs`, `grug.rs`, `blocks.rs`, `server.rs`, `error.rs`, `datetime.rs`.
>     - Examples include changing `format!("{}", var)` to `format!("{var}")`.
>   - **Error Handling**:
>     - Update error message formatting in `error.rs` to use new syntax.
>   - **Logging**:
>     - Update logging statements to use new format string syntax in `server.rs` and `blocks.rs`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=left-curve%2Fleft-curve&utm_source=github&utm_medium=referral)<sup> for 0b6cf82ad7a75c00b38502ac511bbea07175af3b. You can [customize](https://app.ellipsis.dev/left-curve/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/grug/testing/tests/reply.rs
+++ b/grug/testing/tests/reply.rs
@@ -131,7 +131,7 @@ mod replier {
     pub fn reply(ctx: SudoCtx, msg: ReplyMsg, res: SubMsgResult) -> StdResult<Response> {
         let msg = match (res, msg) {
             (GenericResult::Err(wee), ReplyMsg::Fail(execute_msg)) => {
-                println!("replying with error: {}", wee);
+                println!("replying with error: {wee}");
                 execute_msg
             },
             (GenericResult::Ok(_), ReplyMsg::Ok(execute_msg)) => execute_msg,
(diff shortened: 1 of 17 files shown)


grug/testing/tests/reply.rs
```diff
@@ -131,7 +131,7 @@ mod replier {
     pub fn reply(ctx: SudoCtx, msg: ReplyMsg, res: SubMsgResult) -> StdResult<Response> {
         let msg = match (res, msg) {
             (GenericResult::Err(wee), ReplyMsg::Fail(execute_msg)) => {
-                println!("replying with error: {}", wee);
+                println!("replying with error: {wee}");
```
larry0x, line 134: I think this variable should be called `err` instead of `wee`, should be a type, but `wee` sounds funny so I'll keep it.

PR #974, opened 2025-07-29
feat(dango): SingleSigner use Eip712

Rhaki:


dango/client/src/signer.rs
```diff
@@ -55,37 +70,70 @@ where
...
         let username = Username::from_str(username)?;
 
+        let (key, key_hash) = match credential_type {
+            CredentialType::Secp256k1 => (
+                Key::Secp256k1(ByteArray::from_inner(sk.public_key())),
+                sk.public_key().hash256(),
+            ),
+            CredentialType::Ethereum => {
+                let addr =
+                    Addr::from_inner(keccak256(&sk.extended_public_key()[1..])[12..].try_into()?);
```
larry0x, line 88: We have `hyperlane_testing::eth_utils::derive_address` function for this. You can move this whole module to `grug_types` (put under and optional `ethereum` flag) if you need to use it outside of Hyperlane context.

dango/client/src/signer.rs
```diff
@@ -168,12 +216,51 @@ impl Signer for SingleSigner<Defined<u32>> {
...
+                let sign_data = sign_doc.to_sign_data()?;
+
+                Credential::Standard(StandardCredential {
+                    key_hash: self.key_hash,
+                    signature: Signature::Secp256k1(self.sk.sign_digest(sign_data.into()).into()),
+                })
+            },
+            Key::Ethereum(_) => {
+                let verifying_contract =
+                    Some(U160::from_be_bytes(sign_doc.sender.into_inner()).into());
```
larry0x, line 231: We can `impl From<Addr> for alloy::primitive::U160` instead. Put this impl in `grug_types::eth_utils` as well.

dango/client/src/signer.rs
```diff
@@ -180,12 +224,51 @@ impl Signer for SingleSigner<Defined<u32>> {
...
+
+                Credential::Standard(StandardCredential {
+                    key_hash: self.key_hash,
+                    signature: Signature::Eip712(Eip712Signature {
+                        typed_data: data.to_json_vec()?.into(),
+                        sig: sig.into(),
+                    }),
+                })
+            },
+            _ => todo!(),
```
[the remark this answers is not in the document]
larry0x: @Rhaki can you put a sentence here:

> only `Secp256k1` and `Ethereum` key types are supported at this time

dango/client/src/signer.rs
```diff
@@ -1,22 +1,37 @@
...
     std::str::FromStr,
 };
 
 pub const DEFAULT_DERIVATION_PATH: &str = "m/44'/60'/0'/0/0";
 
+#[derive(Debug)]
+pub enum CredentialType {
+    Secp256k1,
+    Ethereum,
+}
```
larry0x, line 33: @Rhaki Instead of introducing this new enum type, I think it's better you create these new functions:

```rust
impl SingleSigner<Undefined<Nonce>> {
    pub fn new_secp256k1(
        username: &str,
        address: Addr,
        sk: SigningKey,
    ) -> anyhow::Result<Self> { /* ... */ }

    pub fn new_ethereum(
        username: &str,
        address: Addr,
        sk: SigningKey,
    ) -> anyhow::Result<Self> { /* ... */ }

    pub fn new_random_secp256k1(
        username: &str,
        address: Addr,
    ) -> anyhow::Result<Self> { /* ... */ }

    pub fn new_random_ethereum(
        username: &str,
        address: Addr,
    ) -> anyhow::Result<Self> { /* ... */ }
}
```

This way, you avoid introducing a new type.

grug/ffi/src/exports.rs
```diff
@@ -35,7 +35,7 @@ pub fn do_instantiate<M, E>(
 ) -> usize
 where
     M: DeserializeOwned,
-    E: Display,
+    E: Debug,
```
larry0x, line 38: @Rhaki can you revert the changes related to error debug here, because it's not the concern of this PR.

grug/vm/rust/src/vm.rs
```diff
@@ -85,6 +85,7 @@ impl Instance for RustInstance {
                     &MockApi,
                     &self.querier,
                 )?;
+
```
larry0x, line 88: Remove this extra line. Let's not touch the code that's not the concern of this PR.

larry0x (changes requested):

dango/client/src/signer.rs
larry0x: Done in [abc4b16](https://github.com/left-curve/left-curve/pull/974/commits/abc4b168e49cbfe050e5436b5b0be309ae4832c8)

PR #979, opened 2025-07-30
[LEF-220] setup codecov in ci

pacmanifold:
> [!IMPORTANT]
> Setup code coverage reporting in CI using Codecov and refactor Pyth handler tests and constants.
> 
>   - **CI/CD**:
>     - Update `.github/workflows/rust.yml` to include code coverage in the test job.
>     - Add `cargo-llvm-cov` installation and run tests with coverage collection.
>     - Upload coverage results to Qlty using `qltysh/qlty-action/coverage@v2`.
>     - Add permissions for `id-token: write` and `contents: read`.
>   - **Testing**:
>     - Refactor `handler()` in `pyth_handler.rs` to use `check_handler_works()` for asserting handler functionality.
>     - Add `create_cache()` test in `client_cache.rs` to verify cache file creation and content.
>   - **Constants**:
>     - Add `PYTH_IDS_ALL` constant in `pyth.rs` for all Pyth IDs.
>     - Rename `CACHE_SAMPLES` to `PYTH_CACHE_SAMPLES` in `client_cache.rs` and update its value to 50.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=left-curve%2Fleft-curve&utm_source=github&utm_medium=referral)<sup> for f01d698e4ce9c2959fb5ae3d96f46037c7d01544. You can [customize](https://app.ellipsis.dev/left-curve/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


dango/testing/tests/pyth_handler.rs
```diff
@@ -66,7 +72,7 @@ fn handler() {
     handler.update_stream(querier, oracle).unwrap();
 
     // Give some times to get the data ready.
-    sleep(Duration::from_millis(500));
+    sleep(Duration::from_millis(3000));
```
penso, line 75: Instead of waiting a specific amount of time, you should reconsider what you're testing. If you're trying to test if the streaming is working, my guess is the current code:

```rust
    // Give some times to get the data ready.
    sleep(Duration::from_millis(3000));

    // Assert the streaming is working.
    for _ in 0..3 {
        assert!(!handler.fetch_latest_vaas().is_empty());
        sleep(Duration::from_millis(500));
    }
```

doesn't necessarily do that. You should check `fetch_latest_vaas()` returns different results. Reading this test, I could think the same results are returned. You probably should do something like (pseudo code):

```
// don't use `while`, so it doesn't loop indefinitely, make it so all loops doesn't go longer than a few seconds
for _ in 0..100 {
 if received_updates; break; 
 sleep(constant from how often pyth should receive updates);
}

let previous_data = ...;
for _ in 0..X {
  let new_data = handler.fetch_latest_vaas();
  assert!(!new_data.is_empty());
  assert!(new_data != previous_data);
  previous_data = new_data;
}
```

this way you'll have faster test, and more resilient.

PR #1047, opened 2025-08-12
new matching algo fixes

apollo-sturdy:


larry0x:
SUMMARY

1. When reading market orders from storage, instead of collecting them into a

   ```rust
   Vec<(Price, MarketOrder)>
   ```

   collect into a 

   ```rust
   BTreeMap<(Price, OrderId), MarketOrder>
   ```

   The `Price` in the map key ensures the orders are sorted by price. The order ID suffix is necessary because there may be multiple orders of the same price.

2. When creating a market order, apply bitwise NOT to the order ID, same as we do for limit orders. This ensures that for market orders of the same price, the older one is prioritized.

3. Fixes the issue that for partially filled or unfilled market orders, the unused deposit isn't refunded.

4. Add `created_at_block_height` field to `MarketOrder`. This ensures that market orders are properly charged trading fee. Since a market order is always created at the current block, it's always charged the taker fee.

PR #1068, opened 2025-08-14
Fix indexer tests after clearing price algorithm change

pacmanifold:
> [!IMPORTANT]
> Update `candles.rs` test assertions to align with new price algorithm changes.
> 
>   - **Test Updates**:
>     - Update assertions in `index_candles_with_real_clickhouse_and_one_minute_interval()` to use `open_price`, `highest_price`, and `lowest_price` instead of `close_price` for `open`, `high`, and `low`.
>     - Modify expected volume values in `index_candles_with_real_clickhouse_and_one_second_interval()` to reflect new algorithm outputs.
>     - Adjust order amounts and expected outcomes in `index_candles_with_both_market_and_limit_orders_one_minute_interval()` to match new clearing price logic.
>   - **Behavior**:
>     - Tests now reflect changes in price calculation logic, ensuring correct behavior under new algorithm.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=left-curve%2Fleft-curve&utm_source=github&utm_medium=referral)<sup> for d1e58a643f80fb4124e4a413b06094124bf58ece. You can [customize](https://app.ellipsis.dev/left-curve/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/dango/testing/tests/indexing/candles.rs
+++ b/dango/testing/tests/indexing/candles.rs
@@ -364,9 +364,8 @@ async fn index_candles_with_both_market_and_limit_orders_one_minute_interval() -
     // Make an empty block at timestamps 41.
     suite.make_empty_block();
 
-    // -------------------------------- block 1 --------------------------------
+    // -------------------------------- First trades (two blocks) --------------------------------
 
-    // Block 1
     // Block time: 61 seconds
     // Place the following orders:
     // - limit, sell, price 100000 USDC per DANGO, size 2 DANGO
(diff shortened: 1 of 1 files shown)


dango/testing/tests/indexing/candles.rs
```diff
@@ -497,14 +528,16 @@ async fn index_candles_with_both_market_and_limit_orders_one_minute_interval() -
         .fetch_all(clickhouse_context.clickhouse_client())
         .await?;
 
-    assert_that!(candle_1m.candles).has_length(1);
+    assert_that!(candle_1m.candles).has_length(2);
```
pacmanifold, line 531: @penso I'm uncertain about this change. I needed to break up the order creation into more txs due to the new market order logic. Is it reasonable that two candles would be indexed then? What is the block time here? I assume 1m means 1 minute candles, is that correct?

penso: `candle_1m` is just the name, but if you look at the query builder it does mention `CandleInterval::OneMinute` so yes.

You shouldn't have 2 candles unless the blocks are created over longer than a minute apart, or unless you have more than a single pair.

larry0x:
already fixed in #1020

PR #1092, opened 2025-08-26
feat(dango): pyth lazer

Zexsor33:


pyth/lazer/src/client.rs
```diff
@@ -181,101 +321,101 @@ impl PythClientTrait for PythClientLazer {
...
+                            }
+
+                            continue;
+                        };
 
-                        // Pull all data available from the receiver.
-                        while !receiver.is_empty() {
-                            let mut buffer = vec![];
-                            receiver.recv_many(&mut buffer, 100).await;
+                        // TODO: Ensure all subscriptions are receiving data each x iterations.
```
apollo-sturdy, line 411: Sounds like a good idea. We can call client.unsubcribe().await and then client.subscribe()?

dango/proposal-preparer/src/pyth_handler.rs
```diff
@@ -127,13 +131,28 @@ where
...
+                    let mut stream = loop{
+
+                        match client.stream(ids.clone()).await {
+                            Ok(stream) => {break stream;}
+                            Err(err) => {
+                                attempts += 1;
+
+                                if attempts < CONNECT_ATTEMPTS {
+                                    error!(error = err.to_string(), "Failed to create Pyth stream; attempts: {attempts}");
+                                    sleep(Duration::from_millis(100)).await;
```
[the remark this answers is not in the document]
Zexsor33: The purpose of the reconnection inside the Handler is to try to reconnect a couple of time just in case something weird happens. So yeah, I think exponential backoff is overkill

pyth/lazer/src/client.rs
Zexsor33: Calling unsubscribe will close all connection. I left a comment because I'm not sure this check is needed. I mean, I could include it just to be safe but theoretically speaking, there is no reason why an endpoint should drop only some connection instead of all.

PR #1106, opened 2025-09-02
feat(ui): native components

kyar1s:


ui/portal/app/src/index.ts
```diff
@@ -1,9 +1,10 @@
 import "../assets/global.css";
 
-import { install } from "react-native-quick-crypto";
```
j0nl1, line 3: 'react-native-quick-crypto' is a shim and have to be loaded as the first thing in the app.

j0nl1 (changes requested):

j0nl1 (approved):

PR #1131, opened 2025-09-14
feat(ui): styles & improvements

kyar1s:


ui/portal/web/src/components/dex/OrderBookOverview.tsx
```diff
@@ -245,12 +250,9 @@ const LiveTrades: React.FC<LiveTradesProps> = ({ base, quote }) => {
...
                 ).slice(0, 10)}
               </p>
-              <p className="text-end z-10">
-                {formatNumber(calculateTradeSize(trade, base.decimals).toFixed(), {
-                  ...formatNumberOptions,
-                  maxSignificantDigits: 10,
-                  maxFractionDigits: 5,
-                }).slice(0, 7)}
+              <p className="text-end z-10 flex gap-1 justify-end">
+                {+size < 0.00001 ? <span>{"<"}</span> : null}
```
j0nl1, line 254: You should use Decimal instead of casting to number because can have overflow issues. I suggest to put the condition before the return such as `const isAmountTooSmall = Decimal(size).lge("0.00001");` then you could only wrap in one condition:

```
{isAmounTooSmall ? <><span>{"< "}</span> size</> : size
```

j0nl1 (changes requested):

j0nl1 (approved):

PR #1195, opened 2025-10-03
infra(grug): Db metrics

Rhaki:


networks/localdango/justfile
```diff
@@ -67,7 +75,7 @@ logs:
 reset:
   rm -rfv ./configs/dango/data
   rm -rfv ./configs/dango/indexer
-  docker volume rm $(docker volume ls -q | grep localdango_)
+  docker volume rm $(docker volume ls -q | grep localdango_ | grep -v localdango_grafana_data)
```
penso, line 78: Don't do this, just do `docker compose down -v` instead of those grep, like I already mentioned in Discord. This is too prone to errors.

PR #1198, opened 2025-10-06
dex: do not collect `FillingOutcome` into a vector

larry0x:
Instead of having two separate steps:

- "Fill orders": loop through `bids` and `asks` vectors, convert to `FillingOutcome`'s, collect into a vector
- "Handle filled": loop through the vector of `FillingOutcome`'s, update state, emit events

We merge them into a single step.

----

> [!IMPORTANT]
> Refactor order filling in `order_filling.rs` to use iterators, update `clear_orders_of_pair` in `cron.rs`, and adjust related metrics and logging.
> 
>   - **Behavior**:
>     - Refactor `fill_orders` in `order_filling.rs` to return an iterator instead of collecting `FillingOutcome` into a vector.
>     - Update `fill_bids` and `fill_asks` in `order_filling.rs` to return iterators.
>     - Modify `clear_orders_of_pair` in `cron.rs` to handle `FillingOutcome` as an iterator.
>   - **Metrics**:
>     - Remove `LABEL_DURATION_HANDLE_FILLED` from `metrics.rs` and related code in `cron.rs`.
>     - Adjust metrics in `clear_orders_of_pair` to reflect iterator-based order filling.
>   - **Misc**:
>     - Minor logging adjustments in `cron.rs` to reflect changes in order handling.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=left-curve%2Fleft-curve&utm_source=github&utm_medium=referral)<sup> for 59323438c536464779a3869e1f0d2b3cdee1cfae. You can [customize](https://app.ellipsis.dev/left-curve/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


dango/dex/src/core/order_filling.rs
```diff
@@ -108,28 +108,23 @@ fn fill_bids(
             fee_base,
             fee_quote,
             clearing_price,
-        });
-
-        if volume.is_zero() {
-            break;
-        }
```
larry0x, line 115: This check is actually unnecessary, because now we only collect orders that are matched (partially or fully) into the `bids` and `asks` vectors. There shouldn't be a case where there is an order in the `bids`/`asks` vector but there's not enough volume to fill it.

dango/dex/src/cron.rs
```diff
@@ -384,6 +384,17 @@ fn clear_orders_of_pair(
 
     #[cfg(feature = "metrics")]
     {
+        let num_trades = bids.len() + asks.len();
```
larry0x, line 387: Previously we use `filling_outcomes.len()`. It should equal the lengths of `bids` and `asks` combined.

dango/dex/src/cron.rs
```diff
@@ -430,81 +441,39 @@ fn clear_orders_of_pair(
...
-    {
-        metrics::counter!(crate::metrics::LABEL_TRADES).increment(filling_outcomes.len() as u64);
-
-        metrics::histogram!(
-            crate::metrics::LABEL_TRADES_PER_BLOCK,
-            "base_denom" => base_denom.to_string(),
-            "quote_denom" => quote_denom.to_string()
-        )
-        .record(filling_outcomes.len() as f64);
-    }
```
larry0x, line 494: This metric is moved above to the "orders matched" part

dango/dex/src/cron.rs
```diff
@@ -697,7 +666,7 @@ fn clear_orders_of_pair(
         }
 
         metrics::histogram!(
-            crate::metrics::LABEL_DURATION_HANDLE_FILLED,
+            crate::metrics::LABEL_DURATION_ORDER_FILLING,
```
larry0x, line 669: The "handle filled" step is merged into the "fill orders" step. @Rhaki the dashboard needs to be updated.

PR #1644, opened 2026-03-20
feat(deploy): add configurable image tag to hyperlane deploy

Zexsor33:
Add optional tag parameter to deploy-hyperlane just recipes, defaulting to "latest". Also add pull_policy: always to both docker-compose files to ensure the correct image is always pulled.


larry0x:
Closing stale PRs. Can revisit later if necessary.

PR #1772, opened 2026-04-08
fix(ui): open interest

kyar1s:


kyar1s:
https://github.com/left-curve/left-curve/pull/1786

PR #1782, opened 2026-04-09
feat(perps): prevent dust positions via `min_position_size`

larry0x:
Rename `PairParam::min_order_size` to `min_position_size` and repurpose the check to validate the resulting position rather than the order itself. This prevents dust positions from three vectors:

1. Tiny orders: rejected if `|current_position + order_size| * price` is non-zero but below `min_position_size`.
2. Near-full closes: reduce-only orders are no longer exempt, so closing almost all of a position is rejected if the remainder would be dust. Full closes (result = 0) are always allowed.
3. Liquidation/ADL: a shared `snap_to_full_close` helper ensures that both `compute_close_schedule` and `execute_adl` snap to full close when the remainder would be below `min_position_size`.


dango/perps/src/trade/submit_order.rs
```diff
@@ -307,10 +308,15 @@ pub(crate) fn _submit_order(
...
 
-    if !reduce_only {
-        check_minimum_order_size(size, oracle_price, pair_param)?;
+    {
+        let current_position = taker_state
+            .positions
+            .get(pair_id)
+            .map(|p| p.size)
+            .unwrap_or_default();
+        check_minimum_position_size(current_position, size, oracle_price, pair_param)?;
```
Zexsor33, line 319: I'm wondering if it will make more sense to move this check after line 337, and ensure `check_minimum_position_size(current_position, fillable_size, oracle_price, pair_param)?;`, in case an user is long 0,799999 and select "reduce only" with a size of 0,8

larry0x: Good point

larry0x:
This PR lags main by too much. Closing and re-doing a new one.

PR #1783, opened 2026-04-09
feat(perps): add `client_order_id` support for limit orders

larry0x:
Allow users to attach an optional `client_order_id` (max 36 chars, alphanumeric/hyphen/underscore) to orders. This enables cancellation before the system-allocated order ID is known — critical for HFT workflows where the user sends a cancel before receiving the submit response.

- Add `client_order_id: Option<String>` to `LimitOrder` and `SubmitOrder`
- Add `CancelOrderRequest::ByClientOrderId` variant
- Add `CLIENT_ORDER_IDS` map in state for (user, client_order_id) → OrderId
- Validate uniqueness among active orders, charset, and length on submit
- Clean up mapping on all removal paths (cancel, fill, liquidation, ADL)
- Include client_order_id in query responses and OrderPersisted event

Note: this is state-breaking (Borsh). Existing `LimitOrder` entries require a chain migration to deserialize with the new field.


larry0x:
Lags behind `main` too much. Closing this and re-do in a new PR.

PR #1836, opened 2026-04-14
chore(deploy): add reset_data param to download-db recipe

Rhaki:
## Summary

- Add `reset_data` parameter (default `"false"`) to the `download-db` just recipe, forwarding it to the `download-db.yml` ansible playbook so data can be wiped before downloading a fresh snapshot.

### Completed
- [x] Verified the justfile syntax is correct

### Remaining
- [ ] Test `just download-db testnet true` against a target host

## Manual QA
None — deploy-only change.


larry0x:
replaced by #1837

PR #1861, opened 2026-04-16
feat: httpd config

j0nl1:


localdango/configs/dango/config/app.toml
```diff
@@ -90,6 +90,27 @@ port = 8080
...
+
+# Maximum time in seconds to wait for the client to send a complete request.
+client_request_timeout_secs = 2
+
+# Maximum time in seconds to wait for the client to disconnect gracefully.
+client_disconnect_timeout_secs = 1
+
+# Maximum number of blocking threads per worker for synchronous operations.
+worker_max_blocking_threads = 8
+
```
penso, line 113: localdango being what it is, I think you can add way lower limits, it's localhost only.

PR #1943, opened 2026-04-25
chore(deploy): standardize ansible via uv; bump pinned roles

larry0x:
Make every team member's ansible runs use the same pinned version installed by uv (root `pyproject.toml`'s `ansible>=13.6.0`), regardless of what's on their PATH or where they invoke from. Fix the broken `just setup` along the way, and bump the three pinned Galaxy roles to current latest.
                                                                                                                                                                   
- **Pin `ANSIBLE_CONFIG`** at the top of `deploy/justfile`: `export ANSIBLE_CONFIG := justfile_directory() + "/ansible.cfg"`. Every recipe shell now loads the deploy-local config explicitly, regardless of cwd or a stray `~/.ansible.cfg`.
- **Rewrite `just setup`**. The old recipe was broken — `source .venv/bin/activate` ran in its own subshell so the activation never persisted, and it created a duplicate `deploy/.venv` separate from the root venv. New recipe is `uv sync` + `uv run ansible-galaxy install -r requirements.yml`. `uv sync` walks up to find the root `pyproject.toml`; `uv run` auto-syncs on subsequent calls, so a missed `setup` no longer means wrong-binary runs.
- **Prefix every ansible CLI call with `uv run`** in `deploy/justfile`: 51 invocations rewritten (45 chained `&& ansible-playbook`, 6 standalone `ansible-vault edit`). Also wrap `./deploy-key.sh` and `./debian-key.sh` (they internally call bare `ansible-vault view`, so the recipes that pipe them into `ssh-add` were also failing without a system-wide ansible).
- **Bump pinned Galaxy roles to current latest**:                                                                                                                
  - `artis3n.tailscale`: `0.5.1` → `v5.0.1` — almost certainly a digit-transposition typo from the original commit on 2025-05-13. `v0.5.1` has never existed in the role's release history (oldest published is `v1.0.0`); latest at that commit date was already `v5.0.1`. Went unnoticed because production hosts had the role cached under `~/.ansible/roles/`. The `v` prefix is required because ansible-galaxy matches against the upstream GitHub tag, and artis3n tags are v-prefixed (the other two roles tag bare semver).                                                                                                                               
  - `MonolithProjects.github_actions_runner`: `1.18.7` → `1.28.0`
  - `geerlingguy.swap`: `1.3.0` → `2.0.0`                                                                                                                        
- **Doc update**: `deploy/AGENTS.md`'s `just setup` line updated to describe the new behavior.

### DIFF
--- a/deploy/requirements.yml
+++ b/deploy/requirements.yml
@@ -7,9 +7,9 @@ collections:
 roles:
   - name: artis3n.tailscale
     src: artis3n.tailscale
-    version: "0.5.1"
+    version: "v5.0.1"
   - name: MonolithProjects.github_actions_runner
     src: MonolithProjects.github_actions_runner
-    version: "1.18.7"
+    version: "1.28.0"
   - name: geerlingguy.swap
-    version: "1.3.0"
+    version: "2.0.0"
--- a/deploy/AGENTS.md
+++ b/deploy/AGENTS.md
@@ -13,7 +13,7 @@ This repository manages infrastructure and app deployments using Ansible. Use th
 
 ## Build, Test, and Development Commands
 
-- `just setup`: Create venv, install Ansible, collections, and roles.
+- `just setup`: Sync the root uv venv (installs the pinned ansible) and install Ansible Galaxy collections/roles. Re-run if collections are added to `requirements.yml`.
 - `just cold-provision` | `just provision`: Run base provisioning (the latter skips `setup` tag).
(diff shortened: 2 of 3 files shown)


deploy/requirements.yml
```diff
@@ -7,9 +7,9 @@ collections:
 roles:
   - name: artis3n.tailscale
     src: artis3n.tailscale
-    version: "0.5.1"
+    version: "v5.0.1"
```
larry0x, line 10: @penso `0.5.1` seems like a typo, should be `5.0.1`

PR #1971, opened 2026-04-30
fix(ui): show one row per perps match in trades tape

larry0x:
Each on-chain perps match emits two `OrderFilled` events sharing a `fillId` — one for the maker, one for the taker. The Trades tab was rendering both, producing pairs of identical-time, identical-size, opposite-side rows that misrepresent each match as two separate trades.

Filter the live `perpsTrades` subscription to the taker side only (`isMaker === false`), keeping pre-v0.16.0 fills (`isMaker === null`) unchanged. This matches the standard convention of one tape row per match, colored by the aggressor's direction.


j0nl1 (approved):

larry0x:
Same fix already done by https://github.com/left-curve/left-curve/pull/1969

PR #1992, opened 2026-05-06
feat(perp): add lot size

Zexsor33:


larry0x:
Too much conflicts. We can revisit this idea later if necessary.

PR #2029, opened 2026-05-11
feat(perps): enforce minimum liquidation value

Zexsor33:
## Summary

  - Add `min_liquidation_value` to the perps global `Param`. During liquidation the close schedule now satisfies two requirements: clearing the maintenance margin deficit AND reaching the
   configured cumulative-notional floor. Iteration stays largest-MM-first and is capped by each position's size, so a floor larger than the user's total notional simply closes everything.
  - Genesis (both `deploy/` and `localdango/`) seeds the floor at **$10** so dust-sized liquidations cannot leave the liquidator under-compensated by the liquidation fee. Setting the
  value to zero disables the floor (deficit-only behavior).
  - Add a chain-upgrade migration in `dango/upgrade` that decodes legacy `Param` bytes via a frozen `LegacyParam` schema (kept in a private `legacy` module) and re-saves them under the
  new schema with the floor backfilled to $10. The migration short-circuits when the new schema already decodes, so an operator-tuned floor is not silently reset.
  - Mirror `minLiquidationValue: string` on `PerpsParam` in the TypeScript SDK.

  ## Validation

  ### Completed

  - [x] `just fmt`
  - [x] `just lint` (clippy `-D warnings` clean across the workspace)
  - [x] `dango-perps` unit tests pass — including 3 new `compute_close_schedule` tests covering: floor extends a small deficit-driven close, floor capped by total position size, floor
  spills across pairs (asserted both with and without the floor on the same fixture)
  - [x] `dango-perps::configure` validation test for negative `min_liquidation_value`
  - [x] `dango-upgrade` unit tests pass — including 2 new tests: migration carries every legacy field across + sets the floor; running the migration twice does not overwrite a hand-tuned
  floor

  ## Manual QA

  1. On a local chain, place an underwater position whose deficit alone would close < $10 of notional. Run `liquidate` — verify the close schedule reaches at least $10 (or the user's full
   position size, whichever is smaller).
  2. Set `min_liquidation_value = 0` via governance `Configure` and repeat — verify deficit-only behavior is restored.
  3. Take a snapshot of a chain on the prior `dango` binary, upgrade to this branch, and verify the `Param` query returns `min_liquidation_value = "10"` with every other field unchanged.


larry0x:
Replaced by #2058

PR #2057, opened 2026-05-19
chore(grug): remove IBC / ICS-23 compatibility

larry0x:
IBC compatibility is no longer in the product roadmap. Strip the optional `ibc` cargo feature in its entirety along with every orphaned artefact:

- delete `grug/jellyfish-merkle/src/ics23.rs` (membership proof) and `grug/db/disk/src/ics23.rs` (non-membership proof)
- remove the `IbcDb` trait from `grug/app/src/traits/db.rs`
- strip the `preimages` RocksDB column family (`CF_NAME_PREIMAGES`, `cf_preimages`, commit-time writes, statistics integration) — it existed solely to back ICS-23 non-membership proofs and was never enabled in production builds
- drop the `ibc` feature, the optional `ics23` dep, and the now orphaned `grug-jmt` regular dep (kept in dev-deps) from `grug-jmt`, `grug-app`, `grug-db-disk` and the workspace
- delete `grug/jellyfish-merkle/spec/` — the Quint/TLA+ formal specification only verified ICS-23 proofs and is now dead
- delete `.github/workflows/quint.yml` — only triggered on changes to the deleted spec directory
- remove the `preimages` row from the column-family table in `book/overview/1-grug-architecture.md`
- scrub stray IBC mentions from doc comments in `tree.rs` and `db.rs`

After this commit, `grep -i "ibc|ics23|ics-23"` across the repo (excluding `libc` false positives and built UI bundles) yields no results.


larry0x:
Replaced by #2059

PR #2089, opened 2026-05-25
test: consolidate testing crates into one

larry0x:
Collapse the five testing-related crates:

- `grug/testing`
- `indexer/testing`
- `hyperlane/testing`
- `dango/testing`
- `dango/mocks/httpd`

into a single crate, `dango/testing`. All tests use the same setup and `TestSuite` abstraction.


grug/app/src/submessage.rs
```diff
@@ -13,7 +13,7 @@ use {
 /// to execute C, which emits a message to execute D... so on.
 ///
 /// Without a limit, this can leads to stack overflow which halts the chain.
-const MAX_MESSAGE_DEPTH: usize = 30;
+const MAX_MESSAGE_DEPTH: usize = 10;
```
larry0x, line 16: This is the only behavior change of this PR. We discovered that 30 is actually too big to prevent stack overflow. Reducing it to 10 works.

PR #2114, opened 2026-06-02
feat(oracle): support weighted multi-source price feeds

larry0x:
Map each denom to a `Vec<PriceSourceWithWeight>` instead of a single `PriceSource`. `query_price` now returns the weight-normalized mean of the component prices, using the oldest component timestamp and treating the market as in regular session only if every component is. This lets a commodity be priced as a weighted blend of its front- and next-month futures contracts.

Weights are validated at registration (non-empty, strictly positive) so the read path can trust them. Propagate the type change through the messages, proposal-preparer subscription list, genesis constants, and test scaffolding, and add an upgrade migration that wraps each existing production price source into a one-element weighted list.


larry0x:
Replaced by #2128.

PR #2196, opened 2026-06-24
feat(deploy): enable S3 block backup on mainnet hetzner1

Zexsor33:
## Summary

Enables S3 block backup for **mainnet** by setting `s3_enabled: true` on **hetzner1**, mirroring the existing testnet setup on hetzner2. The credentials (SOPS vault), bucket creation (`create_s3_bucket.yml`), and the every-10-min sync cron (`s3_sync_cron.yml`) are all already wired up and gated on this single flag, so no other config change is needed.

This PR also **clarifies the S3 bucket-naming flow**, which was easy to misread as "a new timestamped bucket per deploy":

- The `dango-<network>-<timestamp>` name set in `full-app.yml` is only a **first-deploy default**.
- For system-wide networks (testnet/mainnet), `read_current_deploy.yml` overwrites it with the bucket name read back from the **live deployment's `.env`**, so the bucket stays **stable across redeploys** — a single persistent bucket, no fragmentation.
- Only PR/devnet (non-system-wide) keep the fresh per-deploy timestamped name.
- The misleading task name `"Set system wide cometbft and dango directories and overwrite s3_bucket"` in `prep_after_lock.yml` is corrected — that task never actually set `s3_bucket`; the real overwrite lives in `read_current_deploy.yml`. Comments now point readers to the source of truth.

### Notes / caveats
- Backfill only covers blocks present in **hetzner1's local block cache** — blocks never cached locally cannot be recovered to S3.
- `just deploy-mainnet` targets all four mainnet hosts, but only hetzner1 has the flag, so only it will create the bucket + cron (matches the testnet single-node pattern).

### Completed
- [x] `yamllint -d relaxed` clean on all changed files (no new warnings introduced)
- [x] Verified bucket-stability mechanism by reading `read_current_deploy.yml` (grep-back of `S3_BUCKET` gated on `system_wide_directories and not reset_data`)
- [x] Confirmed the change is a pure superset of testnet's hetzner2 host_vars

### Remaining
- [ ] Run `just deploy-mainnet <dango_tag> <bots_tag>` to apply (production deploy — intentionally not done in this PR)
- [ ] After deploy, confirm bucket creation and that the sync cron is uploading blocks

## Manual QA

None (config-only change; effective on next mainnet deploy).

### DIFF
--- a/deploy/full-app.yml
+++ b/deploy/full-app.yml
@@ -36,6 +36,11 @@
         commit_author: "{{ lookup('pipe', 'git log -1 --pretty=%an') }}"
       run_once: true
 
+    # s3_bucket here is only the first-deploy default. For system-wide networks
+    # (testnet/mainnet), read_current_deploy.yml later overwrites it with the
+    # bucket name stored in the existing deployment's .env, so the bucket stays
+    # stable across redeploys. Only PR/devnet (non-system-wide) keep this fresh
+    # per-deploy timestamped name.
     - name: Set timestamp and deployment name on all hosts
       set_fact:
         deployment_name: "{{ hostvars[ansible_play_hosts[0]]['timestamp'] }}"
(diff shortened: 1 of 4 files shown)


larry0x:
We use batched backup now.

PR #2203, opened 2026-06-29
feat: REST/SSE subscriptions

larry0x:
Add Server-Sent Events streaming endpoints replacing the `full_block`
and `perps_events2` GraphQL subscriptions over plain `text/event-stream`:

- `GET /block/full/stream?since=`
- `GET /perps/events/stream?eventTypes=&pairIds=&users=&orderIds=&clientOrderIds=&since=`


larry0x:
Replaced by #2204

PR #2217, opened 2026-07-03
fix(indexer): serve pre-0.26.0 Configure blocks via frozen legacy wire schema

Rhaki:
## Summary

0.26.0 (#2192) removed the taxman contract and changed `Config` in place (dropped `taxman`, added the gas fields that migrated out of taxman's storage). Borsh is positional, so the seven historical blocks that contain a `Message::Configure` no longer deserialize with the current types:

| network | height | date |
|---|---|---|
| mainnet | 16847666 | Apr 8 |
| mainnet | 18782883 | Apr 20 |
| mainnet | 18786698 | Apr 20 |
| mainnet | 18787415 | Apr 20 |
| testnet | 18563232 | Apr 1 |
| testnet | 23062422 | Apr 20 |
| testnet | 23063329 | Apr 20 |

The indexer-httpd answers `500 — failed to load block file: invalid utf-8 sequence` for them (all mainnet nodes; cometbft has pruned far past these heights, so no live source can serve them), and the archive's backfill is currently stalled at 16,847,660 on mainnet.

This is "Option 1" from the team discussion, implemented **once, at the type level**, so every consumer inherits it from the shared crate — the same policy #2192 itself applied to the outcome types ("retained so that historical, Borsh-serialized cached blocks still deserialize"), extended to the one type that could not be retained:

- **dango-primitives**: new `legacy` module — the frozen v1 shadow of the tx path (`ConfigV1 → MsgConfigureV1 → MessageV1 → TxV1 → BlockV1 → FullBlockV1` + the cache payload), read-only by design, with untagged compat views (`FullBlockCompat`, `CachedBlockCompat`, `TxRefCompat`) and whole-payload try-current-then-v1 borsh decoding (borsh can't version-detect mid-stream). **Bytes on disk are untouched — hashes and signatures stay valid** — and a V1 block serializes to exactly the JSON of its era (`taxman` present, no gas fields): no invented values.
- **disk-saver**: `load_bytes()` splits read+decompress off `load()` so one payload can be attempted against multiple schemas.
- **indexer-cache / indexer-httpd**: `CacheFile::load_from_disk_compat` + the `/block` routes (`info`, `result`, `full`, `full/range`) serve the compat shape.
- **archive**: `BlockData` becomes the compat enum — ingest (WS + range fetch) accepts V1 JSON untagged; the RocksDB store writes the enum's one-byte version tag for new values and falls back to the bare decoder for rows persisted before the tag existed; projection + read-API hydration go through schema-agnostic accessors; the co-located `LocalBlockSource` reads node cache files through the same fallback.

Going forward, the module documents the policy that prevents a recurrence: types embedded in persisted data are **append-only** — add a message variant, never change an existing struct in place.

### Completed
- [x] `legacy` golden-bytes tests: bare V1 bytes are **rejected by the current schema** (reproduces the exact production failure) and decoded by the fallback; JSON output contains `taxman` and no `gas_token`
- [x] tagged (store) and untagged (JSON) roundtrips for both eras
- [x] cache-payload decode + projection to the served `{block, outcome}` shape
- [x] archive store: pre-tag bare values read through the fallback (new test), `put`/`get` roundtrip with the tagged format
- [x] full test suites of the touched crates green (`dango-primitives`, `dango-archive-block-source` 25, `dango-archive-projection` 10, `dango-indexer-cache`, `dango-disk-saver`)
- [x] `cargo check --workspace --all-targets` clean
- [x] `cargo clippy --workspace --all-targets --all-features` zero warnings
- [x] `just fmt`

### Remaining
- [ ] Deploy to one mainnet node and verify `/block/full/range?from=16847661&to=16847680` answers 200 with the real block 16847666
- [ ] Verify the archive backfill resumes past 16,847,660 (mainnet) and through the testnet heights when it reaches them
- [ ] Optional: capture one real legacy block file as a fixture for a golden-file regression test

## Manual QA

1. Deploy the dango image on one mainnet full-app host.
2. `curl 'http://<host>/block/full/range?from=16847661&to=16847680' -H 'Host: api-mainnet.dango.zone'` → 200, 20 blocks; block 16847666's configure message shows the era's config (with `taxman`).
3. Deploy the archive image on hetzner7: the healer resumes automatically (frontier passes 16,847,660 within minutes; `archive_block_fetcher_requests_total{outcome="error"}` stops growing).

### DIFF
--- a/dango/archive/block-source/src/httpd_client.rs
+++ b/dango/archive/block-source/src/httpd_client.rs
@@ -84,8 +84,8 @@ impl HttpdClient {
     /// Open a WebSocket subscription to the node's `fullBlock` channel and return
     /// a stream of fully-assembled [`BlockData`], **always starting at the live
     /// tip** (only blocks newer than the current tip). Each data frame's `data`
-    /// is the node's `{block, outcome}` — a `FullBlock`, which `BlockData` (an
-    /// alias of it) deserializes directly.
+    /// is the node's `{block, outcome}` — which `BlockData` (the untagged
+    /// compat enum) deserializes directly, legacy pre-0.26.0 blocks included.
     ///
     /// There is deliberately **no `since`**. The node serves this feed from a
     /// small in-memory ring, so a `since` below that window fails the
--- a/dango/archive/block-source/src/local/mod.rs
+++ b/dango/archive/block-source/src/local/mod.rs
@@ -180,12 +180,12 @@ impl BlockSource for LocalBlockSource {
             return Ok(None);
         }
 
-        let cache_file = CacheFile::load_from_disk_async(path).await?;
+        // Compat load: the co-located node's cache files have the same seven
+        // legacy pre-0.26.0 blocks as everyone else's.
+        let compat =
+            tokio::task::spawn_blocking(move || CacheFile::load_from_disk_compat(path)).await??;
 
-        Ok(Some(BlockData {
-            block: cache_file.data.block,
-            outcome: cache_file.data.block_outcome,
-        }))
+        Ok(Some(compat.into_full_block()))
     }
 
     fn subscribe(&self) -> broadcast::Receiver<Arc<BlockData>> {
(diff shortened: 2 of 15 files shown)


larry0x:
Replaced by #2218

